PR #8, opened 2021-08-11
Disable colors in logs

adaszko:
The ANSI term sequences mess up Google Cloud logs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,15 +33,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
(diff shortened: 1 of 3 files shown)


cloudhead:
We can do this at the code level with: https://docs.rs/tracing-subscriber/0.2.19/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.with_ansi

Will look into it.

cloudhead:
Ok see #8

cloudhead:
Closed in favor of #8

PR #10, opened 2021-08-19
[org-node] Initial event-subscriber implementation

### LINKED ISSUE #4: Use `eth_subscribe` to get real-time updates
Currently we poll TheGraph every 14s for updates. This would be made more efficient if we could subscribe to a real-time event feed. Unfortunately TheGraph doesn't support this, however Geth does and so providers like Alchemy and Infura (?) have an `eth_subscribe` method that allows for event subscription.

We should add a CLI flag to specify an rpc endpoint (`--ethereum-rpc-url`) which supports `eth_subscribe`.

This could be a local node, or a remote service like Alchemy. The org node should then fetch the current project list on startup, from TheGraph, and then subscribe to further `Anchored` and `Unanchored` events from the rpc node.

When an anchor event is received for a project that isn't tracked, it should track it.
When an unanchor event is received, it should stop tracking it. This however is not yet supported by radicle-link.

cloudhead:
Closes #4

### DIFF
--- a/org-node/src/lib.rs
+++ b/org-node/src/lib.rs
@@ -5,6 +5,10 @@
 //!
 //! The org node can be configured to listen to any number of orgs, or *all*
 //! orgs.
+use ethers::abi::Address;
+use ethers::prelude::*;
+use ethers::providers::{Provider, Ws};
+
 use radicle_daemon::Paths;
 use thiserror::Error;
 
@@ -17,8 +21,6 @@ use std::fs::File;
 use std::io;
 use std::net;
 use std::path::PathBuf;
-use std::thread;
-use std::time;
 
 mod client;
 mod query;
@@ -27,10 +29,6 @@ mod store;
 pub use client::PeerId;
 
 use client::{Client, Urn};
-
-/// Default time to wait between polls of the subgraph.
-/// Approximates Ethereum block time.
-pub const DEFAULT_POLL_INTERVAL: time::Duration = time::Duration::from_secs(14);
 
 /// Org identifier (Ethereum address).
 pub type OrgId = String;
@@ -41,9 +39,9 @@ pub struct Options {
     pub cache: PathBuf,
     pub identity: PathBuf,
     pub bootstrap: Vec<(PeerId, net::SocketAddr)>,
+    pub rpc_url: String,
     pub listen: net::SocketAddr,
     pub subgraph: String,
-    pub poll_interval: time::Duration,
     pub orgs: Vec<OrgId>,
     pub timestamp: Option<u64>,
 }
@@ -118,6 +116,12 @@ pub enum Error {
 
     #[error(transparent)]
     Channel(#[from] mpsc::error::SendError<Urn>),
+
+    #[error(transparent)]
+    FromHex(#[from] rustc_hex::FromHexError),
+
+    #[error(transparent)]
+    Query(#[from] Box<ureq::Error>),
 }
 
 /// Run the Node.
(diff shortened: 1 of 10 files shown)


org-node/src/lib.rs
```diff
@@ -164,53 +179,76 @@ pub fn run(rt: tokio::runtime::Runtime, options: Options) -> Result<(), Error> {
...
-                            continue;
-                        }
-                    };
+    // First get up to speed with existing anchors, before we start listening for events.
+    let projects = query(&options.subgraph, store.state.timestamp, &addresses).map_err(Box::new)?;
+    process_anchors(projects, &mut store, &work)?;
 
-                    tracing::info!(target: "org-node", "Queueing {}", urn);
-                    work.blocking_send(urn)?;
+    // Now launch the event subcriber and listen on events.
```
sebastinez, line 194:
```suggestion
    // Now launch the event subscriber and listen on events.
```

org-node/src/lib.rs
```diff
@@ -164,53 +179,76 @@ pub fn run(rt: tokio::runtime::Runtime, options: Options) -> Result<(), Error> {
...
 
-                    if project.timestamp > store.state.timestamp {
-                        tracing::info!(target: "org-node", "Timestamp = {}", project.timestamp);
-
-                        store.state.timestamp = project.timestamp;
-                        store.write()?;
-                    }
-                }
+    while let Some(event) = events.blocking_recv() {
+        match query(&options.subgraph, store.state.timestamp, &[event.address]) {
```
sebastinez, line 198: It's not something important right now, but if we receive the multihash of the project that is being anchored from the WebSocket Log and can decode it with ethers-rs we could eventually avoid querying the subgraph all together..
Would be interesting in that case to leave a `TODO:` here

org-node/src/main.rs
```diff
@@ -26,6 +26,10 @@ pub struct Options {
     #[argh(option)]
     pub subgraph: String,
 
+    /// JSON-RPC WebSocket URL of Ethereum node (eg. ws://localhost:8545)
+    #[argh(option)]
+    pub rpc_url: String,
+
     /// poll interval for subgraph updates (seconds)
     #[argh(option)]
     pub poll_interval: Option<u64>,
```
sebastinez, line 35: I think we can safely remove the poll_interval since we are no longer polling the subgraph

org-node/src/main.rs
```diff
@@ -65,6 +69,7 @@ impl From<Options> for node::Options {
             cache: other.cache,
             listen: other.listen,
             subgraph: other.subgraph,
+            rpc_url: other.rpc_url,
             poll_interval: other
```
sebastinez, line 73: Remove the `poll_interval` option, since we are no longer using it.

org-node/src/lib.rs
```diff
@@ -41,6 +44,7 @@ pub struct Options {
     pub cache: PathBuf,
     pub identity: PathBuf,
     pub bootstrap: Vec<(PeerId, net::SocketAddr)>,
+    pub rpc_url: String,
     pub listen: net::SocketAddr,
     pub subgraph: String,
     pub poll_interval: time::Duration,
```
sebastinez, line 50: We can here also remove the `poll_interval` option.
Also on line 36 the `DEFAULT_POLL_INTERVAL`

sebastinez (commented):
@cloudhead The node works great, did some tests locally, I think we can safely remove the poll_interval since we are no longer polling in a defined interval.
And we could leave a `TODO:` for future references to eventually avoid querying the subgraph all together.

org-node/src/lib.rs
cloudhead: Hmm that's true! I'll open an issue.

PR #19, opened 2021-08-31
Make logging work with GCP

adaszko:
As a bonus: `RUST_LOG`'s value was being ignored.  It no longer is.

### DIFF
--- a/http-api/Dockerfile
+++ b/http-api/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /usr/src/radicle-client-services
 COPY . .
 
 WORKDIR /usr/src/radicle-client-services/http-api
-RUN cargo install --locked --path .
+RUN cargo install --all-features --locked --path .
 
 # Run
 FROM debian:buster-slim@sha256:c8152821b158dd171b4acf92afb0a58fc2faa179a7e0af8ace358fbe1668e99d
@@ -14,4 +14,4 @@ EXPOSE 8777/tcp
 RUN apt-get update && apt-get install -y libssl1.1 && rm -rf /var/lib/apt/lists/*
 COPY --from=build /usr/local/cargo/bin/radicle-http-api /usr/local/bin/radicle-http-api
 WORKDIR /app/radicle
-ENTRYPOINT ["/usr/local/bin/radicle-http-api", "--root", "/app/radicle/root", "--listen", "0.0.0.0:8777", "--no-color"]
+ENTRYPOINT ["/usr/local/bin/radicle-http-api", "--root", "/app/radicle/root", "--listen", "0.0.0.0:8777"]
--- a/http-api/src/main.rs
+++ b/http-api/src/main.rs
@@ -1,32 +1,12 @@
 use std::net;
 use std::path::PathBuf;
-use std::str::FromStr;
 
 use radicle_http_api as api;
 
 use argh::FromArgs;
 
-pub enum LogFmt {
-    Plain,
+use shared::LogFmt;
 
-    #[cfg(feature = "gcp")]
-    Gcp,
-}
-
-impl FromStr for LogFmt {
-    type Err = &'static str;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s {
-            "plain" => Ok(LogFmt::Plain),
-
-            #[cfg(feature = "gcp")]
-            "gcp" => Ok(LogFmt::Gcp),
-
-            _ => Err("Unrecognized log format"),
-        }
-    }
-}
 /// Radicle HTTP API.
 #[derive(FromArgs)]
 pub struct Options {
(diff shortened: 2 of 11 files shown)


org-node/Cargo.toml
```diff
@@ -23,3 +23,8 @@ tracing-subscriber = "0.2"
 ureq = { version = "2.1.1", features = ["json"] }
 ethers = { version = "0.4" }
 rustc-hex = { version = "2.1" }
+tracing-stackdriver = { git = "https://github.com/adaszko/tracing-stackdriver.git", optional = true }
+
+[features]
+default = []
+gcp = ["tracing-stackdriver"]
```
cloudhead, line 30: Can we name is feature `tracing-stackdriver`? Would be more obvious.

org-node/Cargo.toml
```diff
@@ -23,3 +23,8 @@ tracing-subscriber = "0.2"
 ureq = { version = "2.1.1", features = ["json"] }
 ethers = { version = "0.4" }
 rustc-hex = { version = "2.1" }
+tracing-stackdriver = { git = "https://github.com/adaszko/tracing-stackdriver.git", optional = true }
```
cloudhead, line 26: Why are we depending on your fork?

http-api/Dockerfile
```diff
@@ -5,7 +5,7 @@ WORKDIR /usr/src/radicle-client-services
 COPY . .
 
 WORKDIR /usr/src/radicle-client-services/http-api
-RUN cargo install --locked --path .
+RUN cargo install --all-features --locked --path .
```
cloudhead, line 8: Not sure I'd want this as the default, since it won't work in non-gcp contexts, right?

http-api/src/main.rs
```diff
@@ -51,13 +73,44 @@ impl From<Options> for api::Options {
...
+            let env_filter = tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or(tracing_subscriber::EnvFilter::new("info"));
+            let stackdriver_layer = Stackdriver::default();
+            let subscriber = Registry::default().with(stackdriver_layer);
+            let result = env_filter.with_subscriber(subscriber);
+            tracing::subscriber::set_global_default(result)
+                .expect("Could not set up global logger");
+        }
+    }
+}
```
cloudhead, line 107: Can we create a `logger` module at the top level? This code can then be shared between the two crates.

org-node/Cargo.toml
FintanH: You can't name the feature after a package, unfortunately. Or at least it's always failed for me in the past.

adaszko: The name stackdriver is obsolete, as we've discussed in the previous PR.  I'd argue that what's actually more obvious from the operational standpoint is `gcp` because it doesn't delve into internals of a particular log format, which is just an implementation detail.

org-node/Cargo.toml
adaszko: There's a patch in that fork.  We should pull that repo under the radicle-dev github org but as usual, I don't have the necessary permissions.

http-api/src/main.rs
```diff
@@ -28,9 +50,9 @@ pub struct Options {
     #[argh(option, default = r#"String::from("base16-ocean.dark")"#)]
     pub theme: String,
 
-    /// disable colored log output (default: false)
-    #[argh(switch)]
-    pub no_color: bool,
```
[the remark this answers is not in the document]
adaszko: I don't see what for

http-api/Dockerfile
adaszko: It won't cause any failures.  Logging is going to work as if gcp feature didn't exist.

http-api/src/main.rs
adaszko: Sure

adaszko: Done

org-node/Cargo.toml
cloudhead: > You can't name the feature after a package, unfortunately. Or at least it's always failed for me in the past.

A feature is defined by default for each optional package, so this wouldn't require an explicit feature under `[features]`

org-node/Cargo.toml
cloudhead: Can we contribute the patch upstream instead?

http-api/src/main.rs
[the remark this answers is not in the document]
cloudhead: To prevent colors from ending up in the logs.

shared/src/logging.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[cfg(not(feature = "gcp"))]
+pub fn init_logger(log_format: LogFmt) {
+    set_up_plain_log_format(log_format)
+}
+
+#[cfg(feature = "gcp")]
+pub fn init_logger(log_format: LogFmt) {
+    match log_format {
+        LogFmt::Plain => set_up_plain_log_format(log_format),
```
cloudhead, line 43: If we move the `gcp` feature down here, we  don't need the above two functions

org-node/Cargo.toml
cloudhead: > The name stackdriver is obsolete, as we've discussed in the previous PR. I'd argue that from the operational standpoint, `gcp` is actually more obvious because it doesn't delve into internals of a particular log format, which is just an implementation detail.

The point is more that if a feature is essentially activating an optional dependency, then we can just use the name of the dependency instead, as it's already defined as a feature by cargo.

http-api/Dockerfile
cloudhead: Ah sweet.

org-node/Cargo.toml
adaszko: There are critical issues without replies.  The repo is unmaintained.

org-node/Cargo.toml
adaszko: > The point is more that if a feature is essentially activating an optional dependency, then we can just use the name of the dependency instead, as it's already defined as a feature by cargo.

While it may make more sense from the implementation standpoint, it's less intelligible to someone deploying client services.  The latter group is hopefully going to be orders of magnitude bigger.

http-api/src/main.rs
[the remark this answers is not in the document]
adaszko: I can re-add it but it's been added so that machine-readable logs aren't polluted with terminal codes.  Now that the machine-readable logging needs are served by the `gcp` feature, there's no longer need for the black and white logs.  So the original need is no more.

shared/src/logging.rs
adaszko: Good point 👍

http-api/src/main.rs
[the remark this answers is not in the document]
cloudhead: I'd leave it as it's useful for non-gcp deployments.

org-node/Cargo.toml
cloudhead: Ok, I will give you access to radicle-dev then. Also, I'd move this to the top-level Cargo.toml -- see how other git dependencies are handled there.

org-node/Cargo.toml
cloudhead: Ok sure.

shared/src/logging.rs
cloudhead: the body of `set_up_plain_log_format` can be moved here too now, as it's not used anywhere else :+1:

org-node/Cargo.toml
cloudhead: Done.

shared/src/logging.rs
```diff
@@ -0,0 +1,52 @@
...
+                        .unwrap_or(tracing_subscriber::EnvFilter::new("info")),
+                )
+                .init();
+        }
+
+        #[cfg(feature = "gcp")]
+        LogFmt::Gcp => {
+            use tracing_stackdriver::Stackdriver;
+            use tracing_subscriber::Layer;
+            use tracing_subscriber::{layer::SubscriberExt, Registry};
```
cloudhead, line 42:
```suggestion
            use tracing_subscriber::{layer::SubscriberExt, Registry};
            
```

PR #22, opened 2021-09-07
Git Server

cloudhead:


git-server/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+async fn git_handler(
+    ctx: Context,
+    method: Method,
+    headers: HeaderMap,
+    body: impl Buf,
+    remote: Option<net::SocketAddr>,
+    path: warp::filters::path::Tail,
+    query: String,
+) -> Result<impl Reply, Rejection> {
+    let remote = remote.expect("there is always a remote for HTTP connections");
```
xla, line 73: Why make it an `Option` then?

xla (approved):
Would probably name it `git-http-server` to communicate the protocol. </bikeshed>

🕴 ⭐️ ⚠️ 👰

git-server/src/lib.rs
cloudhead: It's `warp` that gives me an `Option` :<

Claimed reason being: the transport may not have a remote addr :shrug:

cloudhead:
> Would probably name it `git-http-server` to communicate the protocol.

That was actually the name of it for a good 15 minutes. Keeping it a bit more general for now as I may include SSH-related stuff in there too.

git-server/src/lib.rs
cloudhead: I'm guessing it's because they support like, unix domain socket as a transport? idk..

cloudhead: Proof enough that warp isn't an HTTP server?

:trollface:

xla:
> That was actually the name of it for a good 15 minutes. Keeping it a bit more general for now as I may include SSH-related stuff in there too.

`gitd` then?

PR #33, opened 2021-09-17
org-node: Exit when a task fails

cloudhead:
This ensures that if any task fails, we fail the entire process so that
it can be restarted.

### DIFF
--- a/org-node/src/lib.rs
+++ b/org-node/src/lib.rs
@@ -159,24 +159,51 @@ pub fn run(rt: tokio::runtime::Runtime, options: Options) -> anyhow::Result<()> 
     let (work, queue) = mpsc::channel(256);
 
     // Queue of events on orgs.
-    let (update, mut events) = mpsc::channel(256);
+    let (update, events) = mpsc::channel(256);
 
-    rt.spawn(client.run(rt.handle().clone()));
-    rt.spawn(track_projects(handle, queue));
+    let client_task = rt.spawn(client.run(rt.handle().clone()));
+    let track_task = rt.spawn(track_projects(handle, queue));
 
     tracing::info!(target: "org-node", "Listening on {}...", options.listen);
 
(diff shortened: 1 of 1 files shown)


org-node/src/lib.rs
```diff
@@ -159,24 +159,51 @@ pub fn run(rt: tokio::runtime::Runtime, options: Options) -> anyhow::Result<()>
...
-    rt.spawn(track_projects(handle, queue));
+    let client_task = rt.spawn(client.run(rt.handle().clone()));
+    let track_task = rt.spawn(track_projects(handle, queue));
 
     tracing::info!(target: "org-node", "Listening on {}...", options.listen);
 
     // First get up to speed with existing anchors, before we start listening for events.
     let projects = query(&options.subgraph, timestamp, &addresses).map_err(Box::new)?;
-    process_anchors(projects, &work)?;
+    rt.block_on(process_anchors(projects, &work))?;
```
xla, line 171: I don't know the exact side-effects this triggers, but whenever something is expected to complete in some timefram I would wrap it in a timeout. As you are using tokio it's readily available as [timeout](https://docs.rs/tokio/1.11.0/tokio/time/fn.timeout.html).

It's also unclear if the tasks above (client, track) should be spawned at all before this op is completed.

xla (commented):
To contextualise the change it would be beneficial to get some insight what prompted it and if there was observed behaviour which triggered it.

org-node/src/lib.rs
cloudhead: There's kind of no point for this to return until it completes, that's why I haven't added a timeout :shrug:

xla: So it's okay for the process to potentially operate in a partially degraded state indefinitely?

cloudhead: well, there's no way around it basically, if this function returns without completing, it's also in a degraded state

PR #47, opened 2021-10-22
[org-node] add Client::track_peers process for tracking and fetching remote peers

### LINKED ISSUE #46: [org-node] track project peers
The `org-node` currently tracks projects, but not its peers. When the `org-node` discovers a known project with an unknown remote (peer), the remote should be tracked and its version of the project fetched.

Ryanmtate:
Resolves #46

### DIFF
--- a/org-node/src/client.rs
+++ b/org-node/src/client.rs
@@ -5,15 +5,15 @@ use either::Either;
 use futures::{future::FutureExt as _, select, stream::StreamExt as _};
 use thiserror::Error;
 
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, oneshot};
 use tokio_stream::wrappers::ReceiverStream;
 
 use librad::{
     git::{identities, refs, replication, storage, storage::fetcher, tracking},
     net::{
         discovery::{self, Discovery as _},
-        peer::{self, Peer},
-        protocol::{self, membership, PeerInfo},
+        peer::{self, event::upstream::Gossip, Peer, PeerInfo, ProtocolEvent},
+        protocol::{self, broadcast::PutResult::Uninteresting, gossip::Payload, membership},
         Network,
     },
     paths::Paths,
@@ -22,8 +22,7 @@ use librad::{
 use crate::client::handle::Request;
 
 pub use handle::{Handle, TrackProjectError};
-pub use librad::git::identities::Urn;
-pub use librad::PeerId;
+pub use librad::{git::Urn, PeerId};
 pub use signer::Signer;
 
 pub mod handle;
@@ -68,6 +67,17 @@ pub enum Error {
     #[error("sending reply failed for {0}")]
     Reply(String),
 
+    #[error("project head was not found")]
+    ProjectHeadNotFound,
+
+    #[error("project peer failed to be tracked and fetched: {0}")]
+    TrackPeer(String),
+
+    #[error("tokio channel receiver error: {0}")]
+    Recv(#[from] tokio::sync::oneshot::error::RecvError),
+
+    #[error("Join handle error: {0}")]
+    TaskJoin(#[from] tokio::task::JoinError),
     #[error(transparent)]
     RefStorage(#[from] Box<radicle_daemon::git::refs::stored::Error>),
 }
@@ -90,6 +100,10 @@ pub struct Config {
     pub limits: Limits,
     /// Listen address.
     pub listen: SocketAddr,
+    /// Track a list of peers
+    pub peers: Vec<PeerId>,
+    /// Allow tracking unknown peers
+    pub allow_unknown_peers: bool,
 }
 
 impl Default for Config {
@@ -98,6 +112,8 @@ impl Default for Config {
             bootstrap: vec![],
             limits: Default::default(),
             listen: ([0, 0, 0, 0], 0).into(),
+            peers: vec![],
+            allow_unknown_peers: false,
         }
     }
 }
@@ -124,6 +140,19 @@ pub struct Head {
     remote: String,
     branch: String,
 }
+
+/// Project peer contains details used to track the peer for an interested project Urn.
+/// this struct is primarily used by `TrackPeerTransmitter`
+struct TrackPeer {
+    api: Peer<Signer>,
+    peer_info: PeerInfo<std::net::SocketAddr>,
+    paths: Paths,
+    urn: Urn,
+}
+
+/// Project peer struct with a acknowledgement transmitter.
+/// This struct is primarily used by `protocol_listener` and TrackProject request handler.
+type TrackPeerTransmitter = (TrackPeer, oneshot::Sender<Result<Option<PeerId>, Error>>);
 
 /// Client instance.
 pub struct Client {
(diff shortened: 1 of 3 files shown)


org-node/src/client.rs
```diff
@@ -277,70 +523,24 @@ impl Client {
...
-                            // maintainer.
-                            match Client::set_head(&urn, &remote, &branch, paths) {
-                                Err(Error::SetHead(err)) => {
-                                    tracing::error!(target: "org-node", "Error setting head: {}", err);
-                                }
-                                Err(err) => return Err(err),
-                                Ok(()) => {}
-                            }
-                            Client::sign_refs(urn, api).await?;
+                if let Some(peer) = peers.next().await {
```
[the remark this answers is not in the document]
cloudhead: Hmm I don't think that's right: the while loop is there because we are attempting peers one by one until the first one succeeds to replicate. With an `if`, we would only be trying the first peer we find.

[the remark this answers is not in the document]
Ryanmtate: 3b77898 reverts to using while loop; I think what I was missing originally was continuing on an error for the match arm, such that the loop does not return an early early, breaking the loop.

org-node/src/client.rs
```diff
@@ -125,6 +141,19 @@ pub struct Head {
...
 }
 
+/// Project peer contains details used to track the peer for an interested project Urn.
+/// this struct is primarily used by `ProjectPeerTransmitter`
+struct ProjectPeer {
+    pub api: Peer<Signer>,
+    pub peer_info: PeerInfo<std::net::SocketAddr>,
+    pub paths: Paths,
+    pub urn: Urn,
+}
```
cloudhead, line 151: Do the fields need to be public? The struct isn't.

cloudhead: I would also find a better name for this that communicates its intent, eg. `TrackPeer`, since it's basically a command, right?

org-node/src/client.rs
```diff
@@ -235,11 +285,200 @@ impl Client {
...
+            let api = rx.0.api;
+            let urn = rx.0.urn;
+            let peer_info = rx.0.peer_info;
+            let paths = rx.0.paths;
+
+            let ack = rx.1;
+
+            match Client::track_project(&api, &urn, &peer_info).await {
+                Err(err) => {
+                    tracing::error!(target: "org-node", "Error tracking {}: {}", urn, err);
```
cloudhead, line 302: We don't return any error back, so how does the sender know to retry?

org-node/src/client.rs
```diff
@@ -235,11 +285,200 @@ impl Client {
...
+                Ok(Err(e)) => {
+                    tracing::error!(target: "org-node", "Project peer returned error: {}", e);
+                    Err(e)
+                }
+                Err(e) => {
+                    tracing::error!(target: "org-node", "Error project peer receiver failed: {}", e);
+                    Err(Error::Recv(e))
+                }
+                Ok(res) => res,
+            },
```
cloudhead, line 414: Any way to avoid these nested result matches? It's a bit hard to follow.

org-node/src/client.rs
```diff
@@ -231,6 +247,53 @@ impl Client {
         }
     }
 
+    /// Track peers that are found by the node that are interested in shared projects;
+    /// This process is spawned within `Client::run` and is run indefinitely.
+    async fn track_peers(node: Peer<Signer>, known_peers: Vec<PeerId>, allow_unknown_peers: bool) {
+        tracing::info!(target: "org-node", "Spawning Peer Tracking Listener");
+        let events = node.subscribe();
+        pin_mut!(events);
```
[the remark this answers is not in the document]
cloudhead: Well, it's a macro that shadows a variable with another variable, from my understanding, ie. it has some effects on the surrounding code. It was created to patch the deficiencies of the async/await feature of the language.. I can try to see if we can get around it, I just find it very hacky.

org-node/src/client.rs
```diff
@@ -282,69 +521,30 @@ impl Client {
                 let mut peers = api.providers(urn.clone(), timeout);
 
                 while let Some(peer) = peers.next().await {
-                    match Client::track_project(api, &urn, &peer).await {
-                        Err(err) => {
-                            tracing::error!(target: "org-node", "Error tracking {}: {}", urn, err);
+                    // Send notification to track peers process;
+                    let response = match Client::track_peer_handler(
```
cloudhead, line 525: Maybe move this out to make the code a little easier to read:
```rust
let result = Client::track_peer_handler(...).await;
let response = match result { ...
```

org-node/src/client.rs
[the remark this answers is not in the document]
Ryanmtate: Hmm, I thought it is to pin the memory to the stack without needing to write your own unsafe code.

org-node/src/client.rs
Ryanmtate: 6e416d0 renames, 5dfbfd6 removes `pub` fields.

org-node/src/client.rs
```diff
@@ -235,11 +285,200 @@ impl Client {
...
 
+    /// Track peers that are found by the node that are interested in shared projects;
+    /// This process is spawned within `Client::run` and is run indefinitely.
+    async fn track_peers(mut peer_rx: mpsc::Receiver<ProjectPeerTransmitter>) {
+        tracing::info!(target: "org-node", "Spawning Track Peers Listener");
+        while let Some(rx) = peer_rx.recv().await {
+            let api = rx.0.api;
+            let urn = rx.0.urn;
+            let peer_info = rx.0.peer_info;
+            let paths = rx.0.paths;
```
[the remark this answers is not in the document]
Ryanmtate: b80afa3 destructures parameters within the if let binding.

org-node/src/client.rs
Ryanmtate: Thanks, nice catch! 5dfbfd6 adds.

org-node/src/client.rs
Ryanmtate: 5dfbfd6 removes nested match arms

org-node/src/client.rs
Ryanmtate: 5dfbfd6 fixes.

org-node/src/client.rs
```diff
@@ -282,69 +513,33 @@ impl Client {
...
+                    )
+                    .await;
+
+                    // Handle the track peer response
+                    let response = match result {
+                        Ok(response) => response,
+                        Err(e) => {
+                            // QUESTION: Should we reply with an error here or continue?;
+                            tracing::error!(target: "org-node", "Error tracking peers: {:?}", e);
+                            continue;
```
Ryanmtate, line 534: Should we be returning an error here instead of continuing?

```
reply
.send(Err(TrackProjectError::<REASON>))
.map_err(|_| Error::Reply("TrackProject".to_string()))
```

org-node/src/client.rs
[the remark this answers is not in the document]
cloudhead: it is, but it does so by shadowing the var with another var, see: https://docs.rs/pin-utils/0.1.0/src/pin_utils/stack_pin.rs.html#14-25

org-node/src/client.rs
cloudhead: No, we return a `NotFound` error if none of the peers we try succeed. It's okay if some of the attempts fail, we just care if one succeeds.

Ryanmtate: eaa3a24 removes question comment.

org-node/src/client.rs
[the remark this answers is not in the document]
Ryanmtate: 9ba4f7a removes and uses `.for_each`

cloudhead (approved):

PR #53, opened 2021-11-01
[org-node] Listen for `Applied(update)` gossip event and update local head to tracked peers updated head.

Ryanmtate:


Ryanmtate:
Works build off #47

cloudhead:
Don't forget that there is only one `HEAD` per namespace, so we shouldn't update it everytime a peer announces a new ref, as that would allow any tracked peer to overwrite a maintainer's ref; we should only update the HEAD when we receive an announcement from a maintainer.

cloudhead (approved):

PR #54, opened 2021-11-03
Create readonly user and initial dashboard at startup

adaszko:


influxdb/00-init.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
cloudhead, line 1:
```suggestion
#!/bin/sh
```

influxdb/template.yaml
```diff
@@ -0,0 +1,58 @@
...
+                  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
+                  |> filter(fn: (r) => r["_measurement"] == "membership" or r["_measurement"] == "peers")
+                  |> filter(fn: (r) => r["_field"] == "active" or r["_field"] == "connected" or r["_field"] == "passive")
+                  |> filter(fn: (r) => r["peer_id"] == "hyyqpngdoe4x4oto3emfdppbw7sj1pfaghbpmmhz5rqiuqg8uofmeo")
+                  |> aggregateWindow(every: v.windowPeriod, fn: last, createEmpty: false)
+                  |> yield(name: "last")
+        width: 5
+        xCol: _time
+        yCol: _value
+    name: org-node
```
cloudhead, line 58: Newline?

influxdb/00-init.sh
adaszko: I'm using various bash-specific options for shell script robustness:
```
set -o errexit
set -o nounset
set -o pipefail
set -o errtrace
```
They aren't available in regular sh.

influxdb/template.yaml
adaszko: Added

influxdb/00-init.sh
cloudhead: Okk gotcha.

cloudhead:
Rebased and merged in c095d42a5b5c4850c18a97aa6f1d67459e160ea8

Issue #56, opened 2021-11-04
[org-node] Need to announce peer update to the gossip protocol


Ryanmtate:
When the push is received, the ref is announced to the network. Currently, the org-node does not announce updates.

Gossip new ref.

- [ ] Add announcement on startup;
- [x] Add announcement on receive update from `git-server`;

cloudhead:
Example here: https://github.com/radicle-dev/radicle-link/blob/master/daemon/src/peer/gossip.rs

cloudhead:
Announcement on startup example here: https://github.com/radicle-dev/radicle-link/blob/master/daemon/src/peer/announcement.rs

Ryanmtate:
Closed by #58

PR #58, opened 2021-11-05
[org-node] Announce update to network peers after updating references

Ryanmtate:
Implements #56 

Builds off of #53

### DIFF
--- a/org-node/src/client.rs
+++ b/org-node/src/client.rs
@@ -353,20 +353,32 @@ impl Client {
 
                     match Client::get_project_head(&urn, &api).await {
                         Ok(Some(Head { remote, branch })) => {
-                            if let Err(err) = Client::set_head(&urn, &remote, &branch, &paths) {
-                                tracing::error!(target: "org-node", "Error setting head: {}", err);
-                                ack.send(Err(err)).ok();
-                                continue;
-                            }
+                            match Client::set_head(&urn, &remote, &branch, &paths) {
+                                Err(err) => {
+                                    tracing::error!(target: "org-node", "Error setting head: {}", err);
+                                    ack.send(Err(err)).ok();
+                                    continue;
+                                }
+                                Ok(rev) => {
+                                    if let Err(err) = Client::sign_refs(urn.clone(), &api).await {
+                                        tracing::error!(target: "org-node", "Error signing refs: {}", err);
+                                        ack.send(Err(err)).ok();
+                                        continue;
+                                    }
 
-                            if let Err(err) = Client::sign_refs(urn, &api).await {
-                                tracing::error!(target: "org-node", "Error signing refs: {}", err);
-                                ack.send(Err(err)).ok();
-                                continue;
-                            }
+                                    // Announce the peer update to the network after setting the head.
+                                    if let Err(e) = api.announce(Payload {
+                                        urn,
+                                        rev: Some(rev),
+                                        origin: Some(peer_id),
+                                    }) {
+                                        tracing::error!(target: "org-node", "Error announcing ref update: {:?}", e);
+                                    }
 
-                            // Acknowledge success;
-                            ack.send(Ok(response)).ok();
+                                    // Acknowledge success;
+                                    ack.send(Ok(response)).ok();
+                                }
+                            }
                         }
                         Ok(None) => {
                             tracing::error!(target: "org-node", "Project head not found! Cannot set head for peer");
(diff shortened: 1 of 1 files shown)


Ryanmtate:
@cloudhead Think we're ready for review. the announce is only happening in the `track_peer_handler`, want to confirm that this is the only place we need it.

You had mentioned a startup process for updating the refs, which I think may still need to be added, but want to consult with you where it should be implemented.

org-node/src/client.rs
```diff
@@ -359,12 +359,22 @@ impl Client {
...
+                            if let Err(err) = Client::sign_refs(urn.clone(), &api).await {
                                 tracing::error!(target: "org-node", "Error signing refs: {}", err);
                                 ack.send(Err(err)).ok();
                                 continue;
                             }
 
+                            // Announce the peer update to the network.
+                            if let Err(e) = api.announce(Payload {
+                                urn,
+                                // QUESTION: Should the current revision be added here?
```
cloudhead, line 371: Yeah, I would say so. Don't quite understand what happens if you don't include a rev, ie. how peers are supposed to interpret that..

org-node/src/client.rs
```diff
@@ -359,12 +359,22 @@ impl Client {
...
+                                rev: None,
+                                origin: Some(peer_id),
+                            }) {
+                                tracing::error!(target: "org-node", "Error announcing peer update: {:?}", e);
```
cloudhead, line 375:
```suggestion
                                tracing::error!(target: "org-node", "Error announcing ref update: {:?}", e);
```

cloudhead (changes requested):

cloudhead:
> @cloudhead Think we're ready for review. the announce is only happening in the `track_peer_handler`, want to confirm that this is the only place we need it.
> 

Not sure I get the intent here. What exactly are we announcing to the network? We need to announce after a push has happened.

Ryanmtate:
> > @cloudhead Think we're ready for review. the announce is only happening in the `track_peer_handler`, want to confirm that this is the only place we need it.
> 
> Not sure I get the intent here. What exactly are we announcing to the network? We need to announce after a push has happened.

Good point, I think I had a second announce on the Gossip applied update match arm, must have deleted it when handling the merge conflict by accident...

Ryanmtate:
> > @cloudhead Think we're ready for review. the announce is only happening in the `track_peer_handler`, want to confirm that this is the only place we need it.
> 
> Not sure I get the intent here. What exactly are we announcing to the network? We need to announce after a push has happened.

The thinking with `track_peer_handler` is that it calls `set_head`, which I think we would want to announce to the network... unless I am mistaken 🤔

org-node/src/client.rs
```diff
@@ -353,20 +353,32 @@ impl Client {
...
+                                        continue;
+                                    }
+
+                                    // Announce the peer update to the network after setting the head.
+                                    if let Err(e) = api.announce(Payload {
+                                        urn,
+                                        rev: Some(rev),
+                                        origin: Some(peer_id),
+                                    }) {
+                                        tracing::error!(target: "org-node", "Error announcing ref update: {:?}", e);
```
Ryanmtate, line 375: also using debug print here due to error type of `Gossip::Payload`, can log an oid/urn instead if desired.

org-node/src/client.rs
Ryanmtate: removed announce statement

org-node/src/client.rs
Ryanmtate: removed announce statement from protocol events as extraneous

org-node/src/client.rs
```diff
@@ -587,6 +587,21 @@ impl Client {
...
+
+                let oid = reference.target().expect("reference target must exist");
+
+                // Announce the updated refs to the network.
+                if let Err(e) = api.announce(Payload {
+                    urn,
+                    rev: Some(Rev::Git(oid)),
+                    origin: Some(api.peer_id()),
+                }) {
+                    tracing::error!(target: "org-node", "Error announcing refs: {:?}", e);
```
cloudhead, line 602:
```suggestion
                    tracing::error!(target: "org-node", "Error announcing refs: {}", e);
```

org-node/src/client.rs
```diff
@@ -587,6 +587,21 @@ impl Client {
                     )
                     .map_err(|e| Error::SetHead(Box::new(e)))?;
 
+                let reference = repository
+                    .find_reference(local_branch_ref.to_str().unwrap_or_default())
+                    .map_err(|e| Error::SetHead(Box::new(e)))?;
```
cloudhead, line 592: Don't we already have the reference? Like, why are we trying to find it? Is it to make sure it exists?

Ryanmtate: Can return from `reference_symbolic` and remove extraneous lookup.

Ryanmtate: > Don't we already have the reference? Like, why are we trying to find it? Is it to make sure it exists?

Need to announce the `Rev::Git(oid)`

6fe4e73 removes additional lookup

org-node/src/client.rs
Ryanmtate: Error is payload type, so it does not implement display, maybe can use `e.urn.encode_id()` ?

cloudhead: Oh weird, let's leave it like this for now.

cloudhead (approved):

cloudhead:
Merged in 7a996be

PR #60, opened 2021-11-05
Add design document

cloudhead:


Ryanmtate:
Document looks great, nice work! 

Do we want to mention CLIs under `radicle-client-tools` for interacting with services and tertiary features, e.g.` rad-auth-keys`, etc.

PR #61, opened 2021-11-08
[org-node] Web Socket Subscription Service

Ryanmtate:
Based off of #58

### DIFF
--- a/org-node/src/client.rs
+++ b/org-node/src/client.rs
@@ -359,10 +359,20 @@ impl Client {
                                 continue;
                             }
 
-                            if let Err(err) = Client::sign_refs(urn, &api).await {
+                            if let Err(err) = Client::sign_refs(urn.clone(), &api).await {
                                 tracing::error!(target: "org-node", "Error signing refs: {}", err);
                                 ack.send(Err(err)).ok();
                                 continue;
+                            }
+
+                            // Announce the peer update to the network.
+                            if let Err(e) = api.announce(Payload {
+                                urn,
+                                // QUESTION: Should the current revision be added here?
+                                rev: None,
+                                origin: Some(peer_id),
+                            }) {
+                                tracing::error!(target: "org-node", "Error announcing peer update: {:?}", e);
                             }
 
                             // Acknowledge success;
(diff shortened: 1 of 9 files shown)


org-node/src/client/webserver.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    // Ignore incoming messages from the connected peer.
+    #[allow(clippy::redundant_pattern_matching)]
+    while let Some(_) = peer_ws_rx.next().await {
+        // Do nothing.
+    }
+}
+
+/// Serves a warp web server instance with a websocket endpoint for pub/sub events.
+/// Requires a peer transmitter sender to broadcast incoming peer updates.
```
Ryanmtate, line 97: Update comment, uses message bus to establish connect web socket handler.

org-node/src/client.rs
```diff
@@ -149,16 +168,41 @@ pub struct Head {
...
+);
+
+/// General Message Bus for cross-thread communications.
+pub enum MessageBus {
+    TrackPeer(TrackPeerTransmitter),
+    /// Used to inform the message bus a new web socket connection has been established.
+    WebSocketConnection(mpsc::UnboundedSender<String>),
+}
+
+impl ToString for MessageBus {
```
Ryanmtate, line 198: Would be good to discuss how we want to serialize messages; I don't think we can use serde here unless we implement the serialization ourselves, which I am not opposed to doing, just want to think through how the format of the messages should be responded back to client.

org-node/src/client.rs
```diff
@@ -33,6 +38,17 @@ pub use signer::Signer;
...
+pub mod webserver;
+
+/// Global unique connect peer id counter.
+/// All peers should be subscription based read-only. There are [currently] no mutations
+/// over websocket to the org-node. Multiple peers, some with the same peer id, may connected
+/// to the websocket service, therefore, we use an atomic usize instead of PeerId for the peer mapping
+/// to the websocket stream handler.
+static WEBSOCKET_CONNECTION_ID: AtomicUsize = AtomicUsize::new(1);
+
+/// Connected Peers Mapping Unbounded Sender to CONNECTED PEER ID as usize.
```
cloudhead, line 50:
```suggestion
/// Connected Clients Mapping Unbounded Sender to CONNECTED PEER ID as usize.
```

Ryanmtate:
deprecated in favor of #64

PR #62, opened 2021-11-11
org-node: move unix socket to tmp directory

Ryanmtate:


FintanH (commented):
It might be more future-proof to use [temp_dir](https://doc.rust-lang.org/std/env/fn.temp_dir.html) instead of hard-coding the string.

Alternatively, we could use [TempDir](https://docs.rs/tempdir/0.3.7/tempdir/struct.TempDir.html) to create a named temporary directory.

cloudhead (approved):

Ryanmtate:
> It might be more future-proof to use [temp_dir](https://doc.rust-lang.org/std/env/fn.temp_dir.html) instead of hard-coding the string.
> 
> Alternatively, we could use [TempDir](https://docs.rs/tempdir/0.3.7/tempdir/struct.TempDir.html) to create a named temporary directory.

Good call, ffc2e59 uses `temp_dir` over the hard coding. `TempDir` looks great too! Not opposed to adding another dependency.

cloudhead:
Merged in cdb32c2

FintanH:
> Good call, ffc2e59 uses temp_dir over the hard coding. TempDir looks great too! Not opposed to adding another dependency.

Up to you peeps :) Seems like the previous code is merged now :sweat_smile:

PR #63, opened 2021-11-12
org-node: unlock encrypted key

FintanH:
This change allows an identity file to be specified that is encrypted
and stored using the FileStorage found in radicle-keystore.

It adds an optional parameter `--identity-passphrase` which expects the
passphrase that decrypts the file to get the signing key.

Signed-off-by: Fintan Halpenny <fintan.halpenny@gmail.com>

### DIFF
--- a/org-node/src/client/signer.rs
+++ b/org-node/src/client/signer.rs
@@ -8,6 +8,12 @@ pub struct Signer {
     pub(super) key: SecretKey,
 }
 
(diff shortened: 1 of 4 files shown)


cloudhead (commented):
Looks good! Could you just run `fmt` and fix any build error?

cloudhead:
Merged, thanks!

PR #65, opened 2021-11-16
http-api: add peer id endpoint

Ryanmtate:
Add endpoint to http-api to programmatically return peer id.


Ryanmtate:
Move this to http-api

Ryanmtate:
deprecated in favor of #67

PR #67, opened 2021-11-16
http-api: add route to return peer information, e.g. id

Ryanmtate:
Add `/v1/peer` route to return information about the peer, currently only the `id` of the peer.

Signed-off-by: Ryan <ryan.michael.tate@gmail.com>

### DIFF
--- a/http-api/src/lib.rs
+++ b/http-api/src/lib.rs
@@ -46,18 +46,33 @@ pub struct Context {
 /// Run the HTTP API.
 pub async fn run(options: Options) {
     let paths = Paths::from_root(options.root).unwrap();
+    let storage = ReadOnly::open(&paths).expect("failed to read storage paths");
+    let peer_id = storage.peer_id().to_owned();
+
     let ctx = Context {
         paths,
         theme: options.theme,
     };
-    let api = path("v1")
-        .and(path("projects"))
+
+    let v1 = warp::path("v1");
+
+    let peer = path("peer")
+        .and(warp::get())
+        .and(path::end())
+        .and_then(move || peer_handler(peer_id));
+
+    let projects = path("projects")
         .and(filters(ctx))
-        .or(warp::get().and(path::end()).and_then(root_handler))
+        .or(warp::get().and(path::end()).and_then(root_handler));
+
+    let routes = v1
+        .and(peer)
+        .or(projects)
         .recover(recover)
         .with(warp::cors().allow_any_origin())
         .with(warp::log("http::api"));
-    let server = warp::serve(api);
+
+    let server = warp::serve(routes);
 
     if let (Some(cert), Some(key)) = (options.tls_cert, options.tls_key) {
         server
@@ -69,6 +84,15 @@ pub async fn run(options: Options) {
     } else {
         server.run(options.listen).await
     }
+}
+
+/// Return the peer id for the node identity.
+/// `GET /v1/peer`
+async fn peer_handler(peer_id: PeerId) -> Result<impl warp::Reply, warp::Rejection> {
+    let response = json!({
+        "id": peer_id.to_string(),
+    });
+    Ok(warp::reply::json(&response))
 }
 
 async fn recover(err: Rejection) -> Result<impl Reply, std::convert::Infallible> {
@@ -162,6 +186,11 @@ async fn root_handler() -> Result<impl Reply, Rejection> {
             {
                 "href": "/v1/projects",
                 "rel": "projects",
+                "type": "GET"
+            },
+            {
+                "href": "/v1/peer",
+                "rel": "peer",
                 "type": "GET"
             }
         ]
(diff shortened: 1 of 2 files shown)


Ryanmtate:
Re-uses recents works by @FintanH in the org-node, adding `identity` and `identity_passphrase` options, used for parsing peer id from public key.

@cloudhead You may know if there is an easier way to return the peer id, but wasn't immediately obvious. Happy to update if there is a better way.

cloudhead:
I don't know of another way actually, but it would make sense to move this identity parsing in the `shared` crate at the top-level, and re-use it in both places.

FintanH (commented):
It's a shame this code gets repeated instead of re-used from somewhere. And getting the `PeerId` is something [`rad-profile` supplies](https://github.com/radicle-dev/radicle-link/blob/master/rad-profile/src/lib.rs#L115-L125) in its library. 

I would be interested in seeing if we could consume more of the `rad-profile` and `rad-identities` functionality for your use cases :)

Ryanmtate:
Thanks @cloudhead @FintanH, commit d91da9f removes getting the peer id from the identity/signer

http-api/src/lib.rs
```diff
@@ -46,14 +46,25 @@ pub struct Context {
...
 pub async fn run(options: Options) {
     let paths = Paths::from_root(options.root).unwrap();
+    let storage = ReadOnly::open(&paths).expect("failed to read storage paths");
+    let peer_id = storage.peer_id().to_owned();
+
     let ctx = Context {
         paths,
         theme: options.theme,
     };
+    // let peer_id = peer_id_from_identity(options.identity, options.identity_passphrase)
```
FintanH, line 56: Some stray comment here :)

FintanH (commented):
LGTM :)

http-api/src/lib.rs
```diff
@@ -46,14 +46,25 @@ pub struct Context {
...
+
+    let peer_id_route = warp::path("peer-id")
+        .and(warp::get())
+        .and_then(move || get_peer_id(peer_id));
+
     let api = path("v1")
         .and(path("projects"))
         .and(filters(ctx))
         .or(warp::get().and(path::end()).and_then(root_handler))
+        .or(peer_id_route)
```
cloudhead, line 67: Isn't this going to put it under `/v1/projects/peer-id`?

cloudhead:
For future-proofing, I'd prefer if we returned an object like `{"id": "abcd123"}` from  `GET /v1/peer`. This would allow adding fields in the future without breaking clients.

http-api/src/lib.rs
Ryanmtate: commit b7358c9 removes.

http-api/src/lib.rs
Ryanmtate: thanks, good catch. Commit b7358c9 re-organizes. Although want to confirm that `/v1/projects` should return 404. Seems to be doing this originally in master, but wanted to double check that this is indeed the expected case.

Ryanmtate:
> For future-proofing, I'd prefer if we returned an object like `{"id": "abcd123"}` from `GET /v1/peer`. This would allow adding fields in the future without breaking clients or requiring two requests.

Commit b7358c9 returns json

http-api/src/lib.rs
```diff
@@ -71,6 +85,15 @@ pub async fn run(options: Options) {
...
 
+/// Return the peer id for the node identity.
+/// `GET /v1/peer`
+async fn get_peer_id(peer_id: PeerId) -> Result<impl warp::Reply, warp::Rejection> {
+    let response = json!({
+        "id": peer_id.to_string(),
+    });
+    Ok(warp::reply::json(&response))
+}
+
```
cloudhead, line 96: Can we bring some consistency to the names of things?

peer_id_route -> peer
get_peer_id -> peer_handler

http-api/src/lib.rs
```diff
@@ -46,18 +46,32 @@ pub struct Context {
...
         paths,
         theme: options.theme,
     };
-    let api = path("v1")
-        .and(path("projects"))
+
+    let v1 = warp::path("v1");
+
+    let peer_id_route = path("peer")
+        .and(warp::get())
```
cloudhead, line 60: We probably want `.and(path::end())`, otherwise this will also match `/v1/peer/lol`. Yeah, `warp` has great defaults /s

Ryanmtate: yes, let me add.

http-api/src/lib.rs
Ryanmtate: commit 801ffec renames.

http-api/src/lib.rs
Ryanmtate: fixed in 7ed2c84

cloudhead (approved):
Looks good! Can you rebase/squash?

cloudhead:
Merged in 0cf5628

Issue #72, opened 2021-11-23
[git-server] Support project aliases


cloudhead:
Currently, pushing through the git bridge requires knowing the project id.

It would be nice if the bridge also supported pushing via the project's name, as specified in the project identity, eg.
```
git push https://seed.alt-clients.radicle.xyz/radicle-client-tools.git
```

Note that project names are not unique, so it's possible for there to be duplicates. This should be handled somehow, perhaps by not supporting aliases for projects with non-unique names.

xphoniex:
How about keeping alias mappings in a flat file, read by both `git-server` and `http-api`?

> Note that project names are not unique, so it's possible for there to be duplicates.

Can you explain a case where there are duplicates for the same name?

PR #74, opened 2021-11-26
git-server: add support for project alias

xphoniex:
Closes https://github.com/radicle-dev/radicle-client-services/issues/72

### DIFF
--- a/git-server/Cargo.toml
+++ b/git-server/Cargo.toml
@@ -28,6 +28,7 @@ tokio = { version = "1.2", features = ["macros", "rt", "rt-multi-thread", "sync"
 tracing = "0.1"
 tracing-subscriber = "0.2"
 warp = { version = "0.3.1", features = ["tls"] }
+radicle-source = { version = "0.2.0" }
 
 # hooks feature enabled dependencies
 envconfig = { version = "0.10.0", optional = true }
--- a/git-server/src/lib.rs
+++ b/git-server/src/lib.rs
@@ -10,9 +10,11 @@ use std::convert::TryFrom;
 use std::io::{BufRead, Read};
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
+use std::sync::{Arc, RwLock};
 use std::{io, net};
 
 use http::{HeaderMap, Method};
+use librad::git::Urn;
 use librad::profile::Profile;
 use rand::{distributions::Alphanumeric, thread_rng, Rng};
 use warp::hyper::StatusCode;
@@ -20,6 +22,8 @@ use warp::reply;
 use warp::{self, path, Buf, Filter, Rejection, Reply};
 
 use error::Error;
+
+use radicle_source::surf::vcs::git::namespace::Namespace;
 
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
 
@@ -42,6 +46,7 @@ pub struct Context {
     authorized_keys: Vec<String>,
     cert_nonce_seed: Option<String>,
     allow_unauthorized_keys: bool,
+    aliases: Arc<RwLock<HashMap<String, Namespace>>>,
 }
 
 impl TryFrom<&Options> for Context {
@@ -65,6 +70,7 @@ impl TryFrom<&Options> for Context {
             authorized_keys: options.authorized_keys.clone(),
             cert_nonce_seed: options.cert_nonce_seed.clone(),
             allow_unauthorized_keys: options.allow_unauthorized_keys,
+            aliases: Default::default(),
         })
     }
 }
@@ -106,11 +112,39 @@ impl Context {
 
         Ok(())
     }
+
+    /// Populates alias map with unique projects' names and their urns
+    fn populate_aliases(&self) -> Result<(), Error> {
+        use librad::git::identities;
+        use librad::git::identities::SomeIdentity::Project;
+        use librad::git::storage::read::ReadOnly;
+        use librad::paths::Paths;
+
+        let mut map = self.aliases.write().unwrap();
+        let paths = Paths::from_root(&self.root)?;
+        let storage = ReadOnly::open(&paths)?;
+        let identities = identities::any::list(&storage)?;
+        for identity in identities.flatten() {
+            if let Project(project) = identity {
+                let urn = Namespace::try_from(project.urn().encode_id().as_str()).unwrap();
+                let name = (&project.payload().subject.name).to_string() + ".git";
+                tracing::info!("alias {:?} for {:?}", name, urn.to_string());
+                if let std::collections::hash_map::Entry::Vacant(e) = map.entry(name.clone()) {
+                    e.insert(urn);
+                } else {
+                    tracing::warn!("alias {:?} exists, skipping", name);
+                }
+            }
+        }
+
+        Ok(())
+    }
 }
 
 /// Run the Git Server.
 pub async fn run(options: Options) {
     let ctx = Context::try_from(&options).expect("Failed to create context from service options");
+    ctx.populate_aliases().expect("Failed to populate aliases");
 
     let git_version = Command::new("git")
         .arg("version")
@@ -182,11 +216,30 @@ async fn git_handler(
     headers: HeaderMap,
     body: impl Buf,
     remote: Option<net::SocketAddr>,
-    namespace: String,
+    mut namespace: String,
     request: warp::filters::path::Tail,
     query: String,
 ) -> Result<Box<dyn Reply>, Rejection> {
     let remote = remote.expect("There is always a remote for HTTP connections");
+    if namespace.ends_with(".git") {
+        let ns_id = namespace.strip_suffix(".git").unwrap();
+        if Urn::try_from_id(ns_id).is_ok() {
+            namespace = ns_id.to_string();
+        } else {
+            // not a project-id, thus potentially an alias
+            // if the alias does not exist, rebuild the cache
+            if !ctx.aliases.read().unwrap().contains_key(&namespace) {
+                ctx.populate_aliases()?;
+            }
+            namespace = ctx
+                .aliases
+                .read()
+                .unwrap()
+                .get(&namespace)
+                .map(Namespace::to_string)
+                .unwrap_or(namespace);
+        }
+    }
     let (status, headers, body) = git(
         ctx, method, headers, body, remote, namespace, request, query,
     )?;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4037,6 +4037,7 @@ dependencies = [
  "http",
  "librad",
  "pgp",
+ "radicle-source",
  "rand 0.8.4",
  "shared",
  "thiserror",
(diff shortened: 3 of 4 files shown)


xphoniex:
re listening for new projects, I think we can standardize the alias format (e.g. ending in ".git") and re-fill our mapping on cache misses. We don't expect to get too many misses anyway, and standardizing format helps us only focus on cases that matter, so we should get good performance.

I can add this in, if it's a go.

git-server/Cargo.toml
```diff
@@ -28,6 +28,7 @@ tokio = { version = "1.2", features = ["macros", "rt", "rt-multi-thread", "sync"
 tracing = "0.1"
 tracing-subscriber = "0.2"
 warp = { version = "0.3.1", features = ["tls"] }
+radicle-source = { version = "0.2.0", features = ["syntax"] }
```
cloudhead, line 31: There's no need for the `syntax` feature here, it handles syntax highlighting :)

git-server/src/lib.rs
```diff
@@ -112,6 +113,10 @@ impl Context {
 pub async fn run(options: Options) {
     let ctx = Context::try_from(&options).expect("Failed to create context from service options");
 
+    let mut aliases = HashMap::new();
+    grab_aliases(&options, &mut aliases);
+    let aliases = Arc::new(aliases);
```
cloudhead, line 118: Aliases should go inside `Context`, then `grab_aliases` can be a method on `Context`.

git-server/src/lib.rs
```diff
@@ -385,3 +393,41 @@ fn gen_random_string() -> String {
...
+    let mut duplicates: Vec<String> = Vec::new();
+    let identities = identities::any::list(&storage)
+        .expect("Failed to list identities in storage");
+    for identity in identities {
+        match identity {
+            Ok(Project(project)) => {
+                let urn = Namespace::try_from(project.urn().encode_id().as_str())
+                    .unwrap().to_string();
+                let name = (&project.payload().subject.name).to_string() + ".git";
+                tracing::info!("alias \"{}\" for \"{}\"", name, urn);
```
cloudhead, line 418: If you use `"{:?}"` on a string, it will be quoted.

git-server/src/lib.rs
cloudhead: We should also use `Urn` as the value to make use of the type checker.

cloudhead:
> re listening for new projects, I think we can standardize the alias format (e.g. ending in ".git") and re-fill our mapping on cache misses. We don't expect to get too many misses anyway, and standardizing format helps us only focus on cases that matter, so we should get good performance.
> 
> I can add this in, if it's a go.

Hmm, the thing is I'd like `.git` to be allowed for project IDs but required for aliases. So how about this:

```
if path ends with ".git" {
  if path is a valid project id {
     lookup project
  } else {
    lookup alias
  }
} else {
  lookup project
}
```

git-server/Cargo.toml
xphoniex: Fixed https://github.com/radicle-dev/radicle-client-services/pull/74/files#diff-7150a884e7aeebff16095bb654e6c671a89cbb34d28864e467368c53ea3bff87R31

git-server/src/lib.rs
xphoniex: Fixed https://github.com/radicle-dev/radicle-client-services/pull/74/commits/8e795c8bba5f1e1edb21d3dd531c6a8bea3d75ff#diff-b672a15ad14f54528cebd77aee76d3007962d44bda987b3e0a34597ec80fd6b1R48

git-server/src/lib.rs
xphoniex: Fixed https://github.com/radicle-dev/radicle-client-services/pull/74/commits/8e795c8bba5f1e1edb21d3dd531c6a8bea3d75ff#diff-b672a15ad14f54528cebd77aee76d3007962d44bda987b3e0a34597ec80fd6b1R134

git-server/src/lib.rs
```diff
@@ -385,3 +393,41 @@ fn gen_random_string() -> String {
...
+                }
+                map.insert(name, urn);
+            },
+            _ => {},
+        }
+    }
+
+    for duplicate in duplicates {
+        map.remove(&duplicate);
+    }
```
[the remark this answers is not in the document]
xphoniex: From original [issue](https://github.com/radicle-dev/radicle-client-services/issues/72):
> Note that project names are not unique, so it's possible for there to be duplicates. This should be handled somehow, perhaps by not supporting aliases for projects with non-unique names.

xphoniex:
> Hmm, the thing is I'd like `.git` to be allowed for project IDs but required for aliases. So how about this:
> 
> ```
> if path ends with ".git" {
>   if path is a valid project id {
>     lookup project
>   } else {
>     lookup alias
>   }
> } else {
>   lookup project
> }
> ```
> 
> `<project-id>` -> OK `<project-id>.git` -> OK `<alias>.git` -> -> OK `<alias>` -> ERROR
> 
> The only case where this fails if you use another project's id as your project name, which I don't see why anyone would do laughing

This should be okay now with https://github.com/radicle-dev/radicle-client-services/pull/74/commits/3f5776a431cd2377d9749cc333a8196110d47858. 

And if the remote is set to `http://localhost:8778/hnrkyfoxs5xp45wg1p8q3dkh74qyebqaeunqo.git` we just redirect it to `hnrkyfoxs5xp45wg1p8q3dkh74qyebqaeunqo`.

Though, I noticed non-existant aliases, trigger 3 updates:
```
Nov 30 12:54:49.807  INFO radicle_git_server: version 0.1.0-8e795c8
Nov 30 12:54:49.923  INFO radicle_git_server: alias "delete-this.git" for "hnrkyfoxs5xp45wg1p8q3dkh74qyebqaeunqo"
Nov 30 12:54:50.084  INFO radicle_git_server: git version 2.28.0
Nov 30 12:54:50.088  INFO Server::run{addr=0.0.0.0:8778}: warp::server: listening on http://0.0.0.0:8778
doesnt exist... rebuilding cache
Nov 30 12:54:51.583  INFO radicle_git_server: alias "delete-this.git" for "hnrkyfoxs5xp45wg1p8q3dkh74qyebqaeunqo"
rebuilt cache
Nov 30 12:54:51.585 ERROR radicle_git_server: unauthorized
Nov 30 12:54:51.586  INFO radicle_git_server: 127.0.0.1:50776 "GET /delete-this2.git/info/refs HTTP/1.1" 401 "-" "git/2.28.0" 6.724034ms
doesnt exist... rebuilding cache
Nov 30 12:54:53.091  INFO radicle_git_server: alias "delete-this.git" for "hnrkyfoxs5xp45wg1p8q3dkh74qyebqaeunqo"
rebuilt cache
Nov 30 12:54:53.415  INFO radicle_git_server: 127.0.0.1:50776 "GET /delete-this2.git/info/refs HTTP/1.1" 200 "-" "git/2.28.0" 340.153381ms
doesnt exist... rebuilding cache
Nov 30 12:54:53.768  INFO radicle_git_server: alias "delete-this.git" for "hnrkyfoxs5xp45wg1p8q3dkh74qyebqaeunqo"
rebuilt cache
Nov 30 12:54:54.267  INFO radicle_git_server: 127.0.0.1:50776 "POST /delete-this2.git/git-receive-pack HTTP/1.1" 200 "-" "git/2.28.0" 513.537247ms
```

Twice on `/info/refs`, once before authentication and once after. And on `/git-receive-pack`. This should however be ok for now.

git-server/src/lib.rs
```diff
@@ -182,11 +229,30 @@ async fn git_handler(
     headers: HeaderMap,
     body: impl Buf,
     remote: Option<net::SocketAddr>,
-    namespace: String,
+    mut namespace: String,
     request: warp::filters::path::Tail,
     query: String,
 ) -> Result<Box<dyn Reply>, Rejection> {
     let remote = remote.expect("There is always a remote for HTTP connections");
+    if namespace[namespace.len() - 4..].eq(".git") {
```
cloudhead, line 237: Perhaps you can use `String::ends_with` here

git-server/src/lib.rs
```diff
@@ -182,11 +229,30 @@ async fn git_handler(
...
+        let ns_id = &namespace[..namespace.len() - 4];
```
cloudhead, line 238: And `strip_suffix(".git")` here

git-server/src/lib.rs
```diff
@@ -106,11 +112,52 @@ impl Context {
 
         Ok(())
     }
+
+    /// Grabs projects names and uses projects' names as alias for urns
+    fn grab_aliases(&self) -> Result<(), Error> {
```
cloudhead, line 117: Can we call this something more descriptive, like `populate_aliases` or `load_aliases`?

git-server/src/lib.rs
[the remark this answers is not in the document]
cloudhead: Ah I see. I think we're maybe doing too much for the user here. How about we don't clear the cache everytime, and if there's a duplicate, we just don't overwrite the alias in the cache and log a warning to the user.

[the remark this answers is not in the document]
cloudhead: Something like:
```
if map.contains(name) {
   warn!("duplicate project name {}", name)
} else {
   map.insert(name, urn)
}
```

git-server/src/lib.rs
xphoniex: Fixed https://github.com/radicle-dev/radicle-client-services/pull/74/commits/67c4936ad316ef7490d2c15d4b88ac055b09ffe9#diff-b672a15ad14f54528cebd77aee76d3007962d44bda987b3e0a34597ec80fd6b1R116-R117

git-server/src/lib.rs
xphoniex: Fixed https://github.com/radicle-dev/radicle-client-services/pull/74/commits/67c4936ad316ef7490d2c15d4b88ac055b09ffe9#diff-b672a15ad14f54528cebd77aee76d3007962d44bda987b3e0a34597ec80fd6b1R229

git-server/src/lib.rs
xphoniex: Fixed https://github.com/radicle-dev/radicle-client-services/pull/74/commits/67c4936ad316ef7490d2c15d4b88ac055b09ffe9#diff-b672a15ad14f54528cebd77aee76d3007962d44bda987b3e0a34597ec80fd6b1R228

git-server/src/lib.rs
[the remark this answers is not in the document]
xphoniex: Fixed https://github.com/radicle-dev/radicle-client-services/pull/74/commits/67c4936ad316ef7490d2c15d4b88ac055b09ffe9#diff-b672a15ad14f54528cebd77aee76d3007962d44bda987b3e0a34597ec80fd6b1L139-R137

cloudhead (approved):
Nice work!

cloudhead:
run `cargo clippy --all` to fix the lint

PR #75, opened 2021-11-27
http-api: Add projects route to api

xphoniex:
Two changes here:

1. Moved `root_handler` to root, that's what users should see when they hit `/`.
2. Added `/v1/projects` handler, which lists all projects e.g:
```json
[{"name":"delete-this","urn":"rad:git:hnrkyfoxs5xp45wg1p8q3dkh74qyebqaeunqo"}]
```

### DIFF
--- a/http-api/src/lib.rs
+++ b/http-api/src/lib.rs
@@ -60,12 +60,11 @@ pub async fn run(options: Options) {
         .and(warp::get().and(path::end()))
         .and_then(move || peer_handler(peer_id));
 
-    let projects = path("projects")
-        .and(filters(ctx))
-        .or(warp::get().and(path::end()).and_then(root_handler));
+    let projects = path("projects").and(filters(ctx));
 
-    let routes = v1
-        .and(peer)
+    let routes = path::end()
+        .and_then(root_handler)
+        .or(v1.and(peer))
         .or(v1.and(projects))
         .recover(recover)
         .with(warp::cors().allow_any_origin())
@@ -119,7 +118,8 @@ async fn recover(err: Rejection) -> Result<impl Reply, std::convert::Infallible>
 
 /// Combination of all source filters.
 fn filters(ctx: Context) -> BoxedFilter<(impl Reply,)> {
-    project_filter(ctx.clone())
+    project_root_filter(ctx.clone())
+        .or(project_filter(ctx.clone()))
         .or(tree_filter(ctx.clone()))
         .or(blob_filter(ctx.clone()))
         .or(readme_filter(ctx))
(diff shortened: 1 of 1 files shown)


http-api/src/lib.rs
```diff
@@ -251,6 +260,34 @@ async fn readme_handler(
...
+
+    let storage = ReadOnly::open(&ctx.paths).unwrap();
+    let projects = identities::any::list(&storage).unwrap()
+        .filter_map(|res| {
+            res.map(|id| match id {
+                SomeIdentity::Project(project) => {
+                    let rad_self = Reference::rad_self(Namespace::from(project.urn()), None);
+                    let name = (&project.payload().subject.name).to_string();
+                    let urn = Urn::try_from(rad_self).ok().unwrap().to_string()
+                        .split('/').next().unwrap().to_string();
```
cloudhead, line 280: If you use the URN from the `link_identities::urn::Urn`, you can just call `urn.encode_id()`

http-api/src/lib.rs
```diff
@@ -251,6 +260,34 @@ async fn readme_handler(
...
+async fn project_root_handler(ctx: Context) -> Result<Json, Rejection> {
+    use radicle_daemon::git::identities::SomeIdentity;
+    #[derive(serde::Serialize)]
+    struct Project {
+        name: String,
+        urn: String,
+    }
+
+    let storage = ReadOnly::open(&ctx.paths).unwrap();
+    let projects = identities::any::list(&storage).unwrap()
```
cloudhead, line 273: These errors should be propagated like in the other handlers.

xphoniex: Fixed https://github.com/radicle-dev/radicle-client-services/pull/75/commits/4e629abc369bc3d005f87bbd99a7dfb65987b450#diff-191ca8c6d151a1a4c5243319a8395c8a17c03c6e257bf4d796b004f19a0bf469L272-R276

http-api/src/lib.rs
xphoniex: Fixed https://github.com/radicle-dev/radicle-client-services/pull/75/commits/4e629abc369bc3d005f87bbd99a7dfb65987b450#diff-191ca8c6d151a1a4c5243319a8395c8a17c03c6e257bf4d796b004f19a0bf469L279-R282

xphoniex: Btw, this is returning:
```
[{"name":"delete-this","urn":"hnrkyfoxs5xp45wg1p8q3dkh74qyebqaeunqo"}]
```
whereas api needs the full `/v1/projects/rad:git:hnrkyfoxs5xp45wg1p8q3dkh74qyebqaeunqo` version, which is what we were getting before.

cloudhead: Oh you're right, my bad! Are we just trying to remove a path eg `/foo`? I don't think there should be any path returned :thinking: 

I think I would just use the `Serialize` instance of `Urn` and include it in `Project` as-is.

cloudhead (approved):

PR #83, opened 2021-12-13
Add specific peer id to route params

sebastinez:
This PR adds a `peer_id` to the routes to obtain the information according to a specific peer id.

Since it's a breaking change, we should implement a migration strategy.


cloudhead (approved):
Nice!

PR #88, opened 2021-12-20
Add remotes routes

sebastinez:
This PR adds 2 new routes `/projects/<urn>/remotes` and `/projects/<urn>/remotes/<peerid>` which shows a list of remotes with replicated identities.
And reverts the peer id argument addition from the source code routes and project info.

It also refactors some querying into helper functions to reduce code duplication.

### DIFF
--- a/http-api/src/error.rs
+++ b/http-api/src/error.rs
@@ -20,6 +20,10 @@ pub enum Error {
     #[error("missing delegations in project")]
     MissingDelegations,
 
+    /// Invalid branch name.
+    #[error("invalid branch name")]
+    BranchName,
+
     /// The entity was not found.
     #[error("entity not found")]
     NotFound,
--- a/http-api/src/lib.rs
+++ b/http-api/src/lib.rs
@@ -2,6 +2,7 @@
 mod error;
 mod project;
 
+use std::collections::HashMap;
 use std::convert::TryFrom as _;
 use std::convert::TryInto as _;
 use std::net;
@@ -13,10 +14,9 @@ use warp::reply;
 use warp::reply::Json;
 use warp::{self, filters::BoxedFilter, path, Filter, Rejection, Reply};
 
-use either::Either;
-
 use radicle_daemon::librad::git::identities;
 use radicle_daemon::librad::git::storage::read::ReadOnly;
+use radicle_daemon::librad::git::tracking;
 use radicle_daemon::librad::git::types::{One, Reference, Single};
 use radicle_daemon::{git::types::Namespace, Paths, PeerId, Urn};
 use radicle_source::surf::file_system::Path;
@@ -122,12 +122,14 @@ fn filters(ctx: Context) -> BoxedFilter<(impl Reply,)> {
         .or(commits_filter(ctx.clone()))
         .or(project_filter(ctx.clone()))
         .or(tree_filter(ctx.clone()))
+        .or(remotes_filter(ctx.clone()))
+        .or(remote_filter(ctx.clone()))
         .or(blob_filter(ctx.clone()))
         .or(readme_filter(ctx))
         .boxed()
 }
 
-/// `GET /:project/blob/:peerid/:sha/:path`
+/// `GET /:project/blob/:sha/:path`
 fn blob_filter(ctx: Context) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     #[derive(serde::Deserialize)]
     struct Query {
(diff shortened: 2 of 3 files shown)


http-api/src/lib.rs
```diff
@@ -247,15 +265,58 @@ async fn blob_handler(
...
+        .filter_map(
+            |branch| match !branch.name.to_string().starts_with("rad/") {
+                true => {
+                    browser
+                        .branch(get_remote_head_branch(&branch.name.to_string(), &peer_id).ok()?)
+                        .unwrap();
+                    let history = browser.get();
+                    Some((branch.name.to_string(), history.first().id.to_string()))
+                }
+                false => None,
```
cloudhead, line 306: maybe an `if` statement here is simpler than a `match`

http-api/src/lib.rs
```diff
@@ -420,3 +463,20 @@ fn project_info(urn: Urn, paths: Paths) -> Result<Info, Error> {
...
+
+fn get_project_metadata(urn: &Urn, paths: &Paths) -> Result<project::Metadata, Error> {
+    let storage = ReadOnly::open(paths)?;
+    let project = identities::project::get(&storage, urn)?.ok_or(Error::NotFound)?;
+    let meta: project::Metadata = project.try_into()?;
+
+    Ok(meta)
+}
+
+fn get_remote_head_branch(branch_name: &str, peer_id: &PeerId) -> Result<git::Branch, Error> {
```
cloudhead, line 475: This can be called `remote_branch`, since "head branch" doesn't really mean anything, and "get" implies some kind of fetching/query. Can we also remove the `Result`?

http-api/src/lib.rs
```diff
@@ -406,10 +452,7 @@ fn project_info(urn: Urn, paths: Paths) -> Result<Info, Error> {
         git::Browser::new_with_namespace(
             &repo,
             &namespace,
-            git::Branch::remote(
-                &format!("heads/{}", meta.default_branch),
-                &remote.default_encoding(),
-            ),
+            get_remote_head_branch(&meta.default_branch, &meta.delegates[0]).unwrap(),
```
cloudhead, line 455: You should keep the error handling with `MissingDelegations` by calling `meta.delegates.first()` and if that returns `None`, return `MissingDelegations` using `Option::ok_or`.

http-api/src/lib.rs
```diff
@@ -247,15 +265,58 @@ async fn blob_handler(
...
-    let reference = Reference::head(Namespace::from(project), peer_id, sha);
-    let commits = browse(reference, ctx.paths, |browser| {
-        radicle_source::commits::<PeerId>(browser, None)
+    let meta = get_project_metadata(&project, &ctx.paths).map_err(Error::from)?;
+
+    let repo = git::Repository::new(ctx.paths.git_dir().to_owned()).map_err(Error::from)?;
+    let mut browser = git::Browser::new_with_namespace(
+        &repo,
+        &git::Namespace::try_from(project.encode_id().as_str()).map_err(Error::from)?,
+        get_remote_head_branch(&meta.default_branch, &peer_id).map_err(Error::from)?,
```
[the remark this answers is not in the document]
sebastinez: If we want to use radicle-surf here, yes.
The issue is that we need to create a `Browser` to be able to call `list_branches` see here
https://docs.rs/radicle-surf/latest/radicle_surf/vcs/struct.Browser.html#examples-8

http-api/src/lib.rs
```diff
@@ -247,15 +265,58 @@ async fn blob_handler(
...
+            name: Some(peer_id.default_encoding()),
+        })
+        .map_err(Error::from)?
+        .iter()
+        .filter_map(
+            |branch| match !branch.name.to_string().starts_with("rad/") {
+                true => {
+                    browser
+                        .branch(get_remote_head_branch(&branch.name.to_string(), &peer_id).ok()?)
+                        .unwrap();
```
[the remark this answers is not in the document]
sebastinez: I'm having an issue with using the `?` operator and/or `map_err` since I get the following
```
the` `?` operator can only be used on `Option`s, not `Result`s, in a closure that returns `Option`
```
The solution I see is to discard the error information with eg. `.ok()?` or `into_ok()` would also be great, but is on nightly..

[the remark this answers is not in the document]
sebastinez: I think I found a good compromise.
There is the following error that doesn't allow me to just use map_err
```
the `?` operator can only be used on `Option`s, not `Result`s, in a closure that returns `Option`
```

http-api/src/error.rs
```diff
@@ -16,7 +16,7 @@ pub enum Error {
     #[error("missing default branch in project")]
     MissingDefaultBranch,
 
-    /// The project does not have delegations.
+    // The project does not have delegations.
```
cloudhead, line 19:
```suggestion
    /// The project does not have delegations.
```

PR #91, opened 2021-12-29
Deployment changes required for replicating via git hooks instead of a org-node p2p replication

adaszko:


git-server/start-up.sh
```diff
@@ -0,0 +1,16 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o errtrace
+
+main () {
+    # The destination is a Docker mounted volume so we postpone copying until the startup of the entire container, when the volume is already mounted.
+    mkdir --parents /app/radicle/git/hooks
+    cp --force /usr/local/bin/pre-receive /app/radicle/git/hooks/pre-receive
```
cloudhead, line 11: Reminder that we will have a post-receive hook soon as well, to copy.

git-server/start-up.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
cloudhead, line 1: Is there a way to avoid this script and have the steps in the dockerfile?

adaszko: When the Dockerfile is interpreted (at image build time), the volume containing radicle work dir isn't mounted yet because volumes are mounted at container startup time (and not at image build time), so there's no place to copy the `pre-receive` hook to.   So the short answer is no I'm afraid.

adaszko: Note that the `bash` process only exists for a brief moment at container startup.  After it executes, it's `exec`'d into the actual git-server process.

cloudhead: Ok got it, then I'd ask that we move this script a little bit out of the way, eg. `git-server/docker/start.sh` so it's obvious that it's meant to be used with docker.

adaszko: Sure.  Done ✅

cloudhead (approved):

PR #96, opened 2022-01-11
Dockerfile: Bump all to bullseye and copy post-receive for git-server

xphoniex:
`org-node` docker image can't currently be run as buster debian has GLIBC 2.28-10, whereas 2.29 is needed by stable rust:

```
/lib/x86_64-linux-gnu/libm.so.6: version 'GLIBC_2.29' not found
```

### DIFF
--- a/git-server/Dockerfile
+++ b/git-server/Dockerfile
@@ -10,7 +10,7 @@ WORKDIR /usr/src/radicle-client-services/git-server
 RUN cargo install --all-features --locked --path .
 
 # Run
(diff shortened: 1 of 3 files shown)


xphoniex:
For `git-server`, `post-receive` binary has also been added to entrypoint script (commit https://github.com/radicle-dev/radicle-client-services/commit/923984f332405327815681b5efbd76cc754453ac) so we need https://github.com/radicle-dev/radicle-client-services/pull/96/commits/963f84028af6c67ffc26360b0f6656e641ca03a3.

PR #110, opened 2022-02-03
http-api: Add project alias fetching

sebastinez:
There is surely a way to make this more compact, or efficient, but pretty happy to have made it work 😃

### DIFF
--- a/http-api/src/lib.rs
+++ b/http-api/src/lib.rs
@@ -8,8 +8,10 @@ use std::convert::TryInto as _;
 use std::net;
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::Arc;
 
 use serde_json::json;
+use tokio::sync::RwLock;
 use warp::hyper::StatusCode;
 use warp::reply;
 use warp::reply::Json;
@@ -43,6 +45,30 @@ pub struct Options {
 pub struct Context {
     paths: Paths,
     theme: String,
+    aliases: Arc<RwLock<HashMap<String, Urn>>>,
+}
+
+impl Context {
+    /// Populates alias map with unique projects' names and their urns
+    async fn populate_aliases(&self, map: &mut HashMap<String, Urn>) -> Result<(), Error> {
+        use radicle_daemon::git::identities::SomeIdentity::Project;
+
+        let storage = ReadOnly::open(&self.paths).expect("failed to open storage");
+        let identities = identities::any::list(&storage)?;
+
+        for identity in identities.flatten() {
+            if let Project(project) = identity {
+                let urn = project.urn();
+                let name = (&project.payload().subject.name).to_string();
+
+                if let std::collections::hash_map::Entry::Vacant(e) = map.entry(name.clone()) {
+                    e.insert(urn);
+                }
+            }
+        }
+
+        Ok(())
+    }
 }
 
 /// Run the HTTP API.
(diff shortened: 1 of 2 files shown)


http-api/src/lib.rs
```diff
@@ -43,6 +45,33 @@ pub struct Options {
...
+
+                tracing::info!("alias {:?} for {:?}", name, urn.to_string());
+
+                if let std::collections::hash_map::Entry::Vacant(e) = map.entry(name.clone()) {
+                    e.insert(urn);
+                } else {
+                    tracing::warn!("alias {:?} exists, skipping", name);
+                }
+            }
+        }
```
cloudhead, line 71: We don't really care about non-project aliases since we only suppor them for projects, hence you can simplify this code by using `identities::project::list`.

http-api/src/lib.rs
```diff
@@ -43,6 +45,33 @@ pub struct Options {
...
     paths: Paths,
     theme: String,
+    aliases: Arc<RwLock<HashMap<String, Urn>>>,
+}
+impl Context {
+    /// Populates alias map with unique projects' names and their urns
+    async fn populate_aliases(&self, map: &mut HashMap<String, Urn>) -> Result<(), Error> {
+        use radicle_daemon::git::identities::SomeIdentity::Project;
+
+        let storage = ReadOnly::open(&self.paths).expect("failed to read storage paths");
```
cloudhead, line 55:
```suggestion
        let storage = ReadOnly::open(&self.paths).expect("failed to open storage");
```

http-api/src/lib.rs
```diff
@@ -475,7 +520,27 @@ async fn project_root_handler(ctx: Context) -> Result<Json, Rejection> {
...
+    }
+    let urn = aliases
+        .get(&namestrings)
+        .cloned()
+        .ok_or_else(warp::reject::not_found)?;
+    tracing::debug!("project alias resolved to {}", urn);
+
     let info = project_info(urn, ctx.paths)?;
 
     Ok(warp::reply::json(&info))
```
cloudhead, line 546: It would be clearer here to call `project_urn_handler`, that way the alias handler is only handling the alias resolution.

cloudhead: I think we can also get rid of the debug statements in here, as it's pretty easy to know what's going on by just calling the route.

http-api/src/lib.rs
```diff
@@ -475,7 +520,27 @@ async fn project_root_handler(ctx: Context) -> Result<Json, Rejection> {
...
 
-async fn project_handler(ctx: Context, urn: Urn) -> Result<Json, Rejection> {
+async fn project_urn_handler(ctx: Context, urn: Urn) -> Result<Json, Rejection> {
+    let info = project_info(urn, ctx.paths)?;
+
+    Ok(warp::reply::json(&info))
+}
+
+async fn project_alias_handler(ctx: Context, name: String) -> Result<Json, Rejection> {
+    let namestrings = name.to_string();
```
cloudhead, line 530: Name is already a string.

http-api/src/lib.rs
sebastinez: Since we currently don't have the `identities::project::list` function, we could maintain this function, stripping the debug statements

cloudhead (approved):

PR #113, opened 2022-02-10
git-server: Remove --git-receive-hook arg

xphoniex:


git-server/src/lib.rs
```diff
@@ -71,10 +70,13 @@ impl Context {
 
         tracing::debug!("Root path set to: {:?}", root_path);
 
+        let root = root_path.canonicalize()?;
+        let git_receive_hook = root.join("hooks/receive-hook");
```
cloudhead, line 74: You sure this works and you don't need to do `root.join("hook").join("receive-hook")`?

xphoniex: works on linux, not sure about other path types though

cloudhead: Might be better to let the path module handle path separation anyway.

xphoniex: Hm... changed it, but I expect it to be smart enough to handle those paths as well. (AFAIK they work on windows too, e.g. `C:\Document\proj/root/hooks/...`, not sure though)

cloudhead (approved):
Looks good!

git-server/src/hooks/post_receive.rs
```diff
@@ -349,12 +349,23 @@ impl PostReceive {
...
+        let child = Command::new(hook)
             .stderr(Stdio::inherit())
             .stdout(Stdio::inherit())
             .stdin(Stdio::piped())
             .spawn()
-            .map_err(Error::CustomHook)?;
+            .map_err(|err| match err.kind() {
+                ErrorKind::NotFound => Error::Io(err),
+                _ => Error::CustomHook(err),
+            });
```
cloudhead, line 360: There's no need to change this, `CustomHook` has the `io::Error` inside of it, so you can just match on it and on the error-kind in the conditional below.

xphoniex: https://github.com/radicle-dev/radicle-client-services/pull/113/commits/dfcf3f0f42da80638c52a349ece6e8a233f6bcaa#diff-bcec911f4891588d789f1bf73e0dadb81401adba787874fd6a3eced15bc552d7R359-R364

cloudhead (approved):

PR #116, opened 2022-02-17
http-api: Add date range to commit history route

sebastinez:
This PR adds the possibility to fetch the commit history for a date range.
It allows to add the `since` and `until` query param in a UNIX Timestamp format, e.g.
`/projects/<urn>/commits?since=1643629969&until=1643829969`

### DIFF
--- a/http-api/src/lib.rs
+++ b/http-api/src/lib.rs
@@ -10,6 +10,8 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::Arc;
 
+use radicle_source::commit::Header;
+use serde::{Deserialize, Serialize};
 use serde_json::json;
 use tokio::sync::RwLock;
 use warp::hyper::StatusCode;
@@ -39,6 +41,13 @@ pub struct Options {
     pub tls_cert: Option<PathBuf>,
     pub tls_key: Option<PathBuf>,
     pub theme: String,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+struct CommitsQueryString {
+    parent: Option<String>,
+    since: Option<i64>,
+    until: Option<i64>,
 }
 
 #[derive(Debug, Clone)]
@@ -214,7 +223,7 @@ fn history_filter(ctx: Context) -> impl Filter<Extract = impl Reply, Error = Rej
         .map(move || ctx.clone())
         .and(path::param::<Urn>())
         .and(path("commits"))
-        .and(query::<HashMap<String, String>>())
+        .and(query::<CommitsQueryString>())
         .and(path::end())
         .and_then(history_handler)
 }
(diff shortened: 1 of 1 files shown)


http-api/src/lib.rs
```diff
@@ -389,22 +398,53 @@ async fn remote_handler(
...
+                    q.committer_time.seconds() < until
+                } else {
+                    // If the commitTime is neither after `since` nor before `until` we include the commit header
+                    true
+                }
+            })
+            .cloned()
+            .collect();
+
+        result.headers = headers.clone();
```
cloudhead, line 443: Cloning here shouldn't be necessary if you get the length from `result.headers`..

http-api/src/lib.rs
```diff
@@ -41,6 +43,13 @@ pub struct Options {
     pub theme: String,
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+struct CommitsQueryString {
+    parent: Option<String>,
```
cloudhead, line 48: Could we use `Option<One>` here? Then we wouldn't need to parse in the handler..

http-api/src/lib.rs
```diff
@@ -389,22 +398,53 @@ async fn remote_handler(
...
-        radicle_source::commits::<PeerId>(browser, None)
+        let mut result = radicle_source::commits::<PeerId>(browser, None)?;
+        let headers: Vec<Header> = result
+            .headers
+            .iter()
+            .filter(|q| {
+                if let (Some(since), Some(until)) = (since, until) {
+                    q.committer_time.seconds() > since && q.committer_time.seconds() < until
+                } else if let Some(since) = since {
+                    q.committer_time.seconds() > since
```
cloudhead, line 432: I'd expect `since` to be inclusive.

http-api/src/lib.rs
sebastinez: Yeah we can use `Option<One>`, but we would still need to parse `meta.head` from a `git2::Oid` to a `One` in that case.
Or we could leave it as `Option<String>` and parse in the header to `One`

http-api/src/lib.rs
```diff
@@ -389,22 +398,53 @@ async fn remote_handler(
...
+
     let commits = browse(reference, ctx.paths, |browser| {
-        radicle_source::commits::<PeerId>(browser, None)
+        let mut result = radicle_source::commits::<PeerId>(browser, None)?;
+        let headers: Vec<Header> = result
+            .headers
+            .iter()
+            .filter(|q| {
+                if let (Some(since), Some(until)) = (since, until) {
+                    q.committer_time.seconds() > since && q.committer_time.seconds() < until
```
cloudhead, line 430:
```suggestion
                    q.committer_time.seconds() >= since && q.committer_time.seconds() < until
```

http-api/src/lib.rs
```diff
@@ -389,22 +398,53 @@ async fn remote_handler(
...
+                } else if let Some(since) = since {
+                    q.committer_time.seconds() >= since
+                } else if let Some(until) = until {
+                    q.committer_time.seconds() < until
+                } else {
+                    // If the commitTime is neither after `since` nor before `until` we include the commit header
```
cloudhead, line 436: Shouldn't this comment say "If neither `since` nor `until` are specified, we include the commi"?

cloudhead (approved):

PR #120, opened 2022-02-21
Upload binaries to a GCS bucket

adaszko:


rudolfs (approved):
Looks good!

.github/workflows/actions.yml
```diff
@@ -23,11 +23,47 @@ jobs:
         with:
           profile: minimal
       - name: Build
-        run: cargo build --verbose --all-features
+        run: cargo build --release --verbose --all-features
```
cloudhead, line 26: This means we are building twice for CI to pass, could we make the build step not affect CI check status maybe?

adaszko: Right, good point.  So we can either:

1. Run tests in release mode (`cargo test --release`) or
2. Run tests as is, in debug mode and only do `cargo build --release` when pushing to the `master` branch.

Which one do you prefer?  I don't see anything wrong with 1. personally and it's going to be quicker to execute overall.

.github/workflows/actions.yml
```diff
@@ -23,11 +23,47 @@ jobs:
...
+        uses: 'google-github-actions/auth@v0'
+        with:
+          workload_identity_provider: 'projects/281042598092/locations/global/workloadIdentityPools/github-actions/providers/google-cloud'
+          service_account: 'github-actions@radicle-services.iam.gserviceaccount.com'
+      - name: Upload binaries to Google Cloud Storage
+        if: ${{ github.ref == 'refs/heads/master' }}
+        uses: 'google-github-actions/upload-cloud-storage@v0'
+        with:
+          path: artifacts
+          destination: radicle-client-services/
```
rudolfs, line 65: Could we maybe store them under `radicle-client-services/main/x86_64-linux/` so it is more clear that they're coming from the `main` branch and what arch they're for?

rudolfs: It could be also useful if we could store a copy the same artefacts by commit sha, so that we can try out a specific version. Something like `radicle-client-services/<COMMIT_SHA>/x86_64-linux/`. And limit the gcp retention to something like a couple of GB.

adaszko: Can we first merge this PR?  It's been tested already and proven to work.  I'd rather extend the scope of a task in a separate PR.

.github/workflows/actions.yml
cloudhead: I would prefer (2) because we don't have any tests currently, so (1) would just make ci slower.

PR #122, opened 2022-02-24
git-server: Make it possible to override some args

cloudhead:
Signed-off-by: Alexis Sellier <self@cloudhead.io>


adaszko (approved):
LGTM

PR #127, opened 2022-03-08
deps: remove radicle-daemon

FintanH:
This patch removes the use of the radicle-daemon package. The only use
of it was to import code from the re-exported librad package. Instead
we just use the librad dependency and import straight from there.

Signed-off-by: Fintan Halpenny <fintan.halpenny@gmail.com>


cloudhead (approved):

cloudhead:
Thanks!

PR #132, opened 2022-03-22
http-api: Add commit signature verification to history handler

sebastinez:
A first implementation of commit signature verification.

Using `git show <commit> --pretty'%GF' --raw` we are able to verify that a commit has a valid signature and return the corresponding fingerprint.
On the other side we obtain all the peers from the seed node and check if they match the commit signature information.

At the moment we only verify ssh signatures that match a valid `Person` later we eventually will verify GPG signatures or third party SSH signatures.


http-api/src/lib.rs
```diff
@@ -76,6 +79,44 @@ impl Context {
...
+        let storage = ReadOnly::open(&self.paths).map_err(Error::from)?;
+        let project = identities::project::get(&storage, &urn)
+            .map_err(Error::Identities)?
+            .ok_or(Error::NotFound)?;
+        let meta: project::Metadata = project.try_into()?;
+
+        let identities = identities::any::list(&storage)?;
+        for identity in identities.flatten() {
+            if let SomeIdentity::Person(person) = identity {
+                if let Some(pk) = person.delegations().iter().next() {
```
cloudhead, line 100: Convert this into a `while let` loop because we want *all* delegations of a person here, ie. if they have multiple keys we want all their keys to point to their identity.

http-api/src/lib.rs
```diff
@@ -740,6 +826,30 @@ fn remote_branch(branch_name: &str, peer_id: &PeerId) -> git::Branch {
...
+
+    // If there is no stdout or if there is some stderr we return None early.
+    if result.is_empty() || !output.stderr.is_empty() {
+        return None;
+    }
+
+    // We match stdout against 'SHA256: which is what git show starts with if there is a ssh fingerprint.
+    match result.starts_with(b"'SHA256:") {
+        true => {
+            let fingerprint: Vec<u8> = result.splice(1..51, Vec::new()).collect();
```
cloudhead, line 846: This seems a bit brittle, can we instead get the first line out of the result? I'd also skip the check for SHA256, it will simply not be found if it's not in the map.

cloudhead: You should be able to do something like `io::BufReader::new(result).lines().next()` to get the first line.

http-api/src/lib.rs
```diff
@@ -76,6 +79,44 @@ impl Context {
...
+        for identity in identities.flatten() {
+            if let SomeIdentity::Person(person) = identity {
+                if let Some(pk) = person.delegations().iter().next() {
+                    let peer_id = &PeerId::from(*pk);
+                    let delegate = meta.delegates.iter().any(|d| d.contains(peer_id));
+                    let peer = Peer {
+                        id: peer_id.to_string(),
+                        name: person.subject().name.to_string(),
+                        delegate,
+                    };
```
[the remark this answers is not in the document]
cloudhead: Actually all peers in this map will have a `Person`, so no need for the `Option`.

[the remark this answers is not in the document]
sebastinez: Yeah that's the reason I introduced the `urn` parameter to the `populate_fingerprints` fn but I see the flaw in the data model now. 😏 
:+1

http-api/src/lib.rs
sebastinez: hmm it seems to be able to do a `while let` over an iterator, one has to return a Result variant..
Another option is to iterate with `for pk in person.delegations().iter()` wdyt? It seems to me that it should do the same and is maybe also more descriptive.

sebastinez:
Okay did a rebase fixed conflicts with the pagination PR merge and addresses all the open issues in the most idiomatic Rust way.

http-api/src/lib.rs
cloudhead: `while let` works with any iterator, but you have to create the iterator separately and then call `next()`. Using a `for in` loop is perfectly fine as well.

http-api/src/lib.rs
```diff
@@ -1,17 +1,19 @@
 #![allow(clippy::if_same_then_else)]
+mod commit;
 mod error;
 mod project;
 
+use serde_json::json;
+use shared::keys;
```
cloudhead, line 7: Can you move these two the external dependency grouping of imports?

We usually do

```
<std dependencies>

<external depdencies>

<internal dependencies>
```

http-api/src/lib.rs
```diff
@@ -78,6 +72,38 @@ impl Context {
...
+        let storage = ReadOnly::open(&self.paths).map_err(Error::from)?;
+        let identities = identities::any::list(&storage)?;
+        for identity in identities.flatten() {
+            if let SomeIdentity::Person(person) = identity {
+                for pk in person.delegations().iter() {
+                    let peer_id = &PeerId::from(*pk);
+                    let fp = keys::to_ssh_fingerprint(peer_id).ok();
+
+                    if let std::collections::hash_map::Entry::Vacant(e) =
+                        map.entry(fp.unwrap_or_default())
```
cloudhead, line 92: Using unwrap_or_default here means that if `to_ssh_fingerprint` fails, we will store an entry under the empty string `""`, which is probably not what we want!

Since this shouldn't fail unless we have a bug, I'd unwrap the result like this: 

```rust
                     let fp = keys::to_ssh_fingerprint(peer_id).expect("Conversion cannot fail");
                     ```

File: http-api/src/lib.rs [hunk 6]
```diff
@@ -750,6 +807,25 @@ fn remote_branch(branch_name: &str, peer_id: &PeerId) -> git::Branch {
...
+    // If there is no stdout or if there is some stderr we return None early.
+    if result.is_empty() || !output.stderr.is_empty() {
+        return None;
+    }
+
+    BufReader::new(result.as_slice())
+        .lines()
+        .next()
+        .map(|line| line.unwrap_or_default())
+}
```
cloudhead, line 827: This function really needs to return a `io::Result<String>`, you can then return an io::Error if there are issues.

http-api/src/lib.rs
```diff
@@ -750,6 +807,25 @@ fn remote_branch(branch_name: &str, peer_id: &PeerId) -> git::Branch {
...
+
+    // If there is no stdout or if there is some stderr we return None early.
+    if result.is_empty() || !output.stderr.is_empty() {
+        return None;
+    }
+
+    BufReader::new(result.as_slice())
+        .lines()
+        .next()
+        .map(|line| line.unwrap_or_default())
```
cloudhead, line 826: This should return an error, not an empty string, if the line cannot be parsed. I wouldn't use `map` here because that'll make it harder to exit on error.

http-api/src/lib.rs
```diff
@@ -750,6 +807,25 @@ fn remote_branch(branch_name: &str, peer_id: &PeerId) -> git::Branch {
...
+        .current_dir(ctx.paths.git_dir()) // We need to place the command execution in the git dir
+        .args(["show", sha1, "--pretty=%GF", "--raw"])
+        .output()
+        .expect("failed to execute git show command");
+    let result = output.stdout;
+
+    // If there is no stdout or if there is some stderr we return None early.
+    if result.is_empty() || !output.stderr.is_empty() {
+        return None;
+    }
```
cloudhead, line 821: Rather than checking output.stderr, check that `output.status` is successful

http-api/src/lib.rs
```diff
@@ -750,6 +807,25 @@ fn remote_branch(branch_name: &str, peer_id: &PeerId) -> git::Branch {
     )
 }
 
+fn commit_hash_to_fingerprint(ctx: Context, sha1: &str) -> Option<String> {
+    let output = Command::new("git")
+        .current_dir(ctx.paths.git_dir()) // We need to place the command execution in the git dir
+        .args(["show", sha1, "--pretty=%GF", "--raw"])
+        .output()
+        .expect("failed to execute git show command");
```
cloudhead, line 815: Should return an error instead of panicking.

http-api/src/lib.rs
```diff
@@ -475,16 +505,43 @@ async fn history_handler(
...
             .skip(page * per_page)
             .take(per_page)
-            .cloned()
+            .map(|header| {
+                let id = commit_hash_to_fingerprint(ctx.clone(), &header.sha1.to_string())
+                    .unwrap_or_default();
+                let peer = fingerprints
+                    .get(&id)
+                    .map(|(id, person)| -> Result<Peer, Error> {
+                        let info = project_info(project.to_owned(), ctx.paths.to_owned())?;
```
cloudhead, line 514: Why are we fetching project info on every commit? It doesn't change.

http-api/src/lib.rs
```diff
@@ -475,16 +505,43 @@ async fn history_handler(
             })
             .skip(page * per_page)
             .take(per_page)
-            .cloned()
+            .map(|header| {
+                let id = commit_hash_to_fingerprint(ctx.clone(), &header.sha1.to_string())
+                    .unwrap_or_default();
```
cloudhead, line 510: This will return `""` on failure, which we don't want. I'd return an error a 500 from the history handler in this case, because it means something is wrong with our git installation or our code.

http-api/src/lib.rs
sebastinez: hmm yeah that was my idea too initially, but `cargo fmt --all --check` wants the `serde_json` and the `shared` crate on top..

http-api/src/lib.rs
sebastinez: Yeah this came up because we needed the project info for checking delegates, and at some point I didn't want to create project info for every `history_handler` call.
But it's better than calling it for every commit that has peer information.
So I'll move it to the beginning of the `history_handler`

cloudhead (approved):

cloudhead (approved):
Nice.

PR #133, opened 2022-03-23
http-api: Add commit history pagination

sebastinez:
This PR adds pagination to the commit history handler.

- Sorts the commits by `committer_time` from newest to oldest commit
- Creates chunks of size defined by query param `per_page` and defaults to 30 commits per page.
- Gets chunk at index defined by query param `page` and defaults to first page if None given.


http-api/src/lib.rs
```diff
@@ -450,8 +466,20 @@ async fn history_handler(
 
     let commits = browse(reference, ctx.paths, |browser| {
         let mut result = radicle_source::commits::<PeerId>(browser, None)?;
-        let headers: Vec<Header> = result
+
+        // To create chunks that make sense we sort the commits from newest to oldest.
+        result
             .headers
+            .sort_by(|a, b| b.committer_time.cmp(&a.committer_time));
```
cloudhead, line 473: Isn't it already sorted?

http-api/src/lib.rs
```diff
@@ -450,8 +466,20 @@ async fn history_handler(
...
+        let paginated_headers: Vec<Vec<Header>> = result
+            .headers
+            .chunks(per_page.unwrap_or_default())
```
cloudhead, line 476: This will return `0` when `None`, because `0` is the default of `usize`

http-api/src/lib.rs
```diff
@@ -450,8 +466,20 @@ async fn history_handler(
...
+            .map(|x| x.to_vec())
+            .collect();
```
cloudhead, line 478: We don't want to actually create all pages if we're only going to get the first page for eg. (imagine a large commit history with 100K+ commits). So instead of using `collect()` and then `get()` we should be able to use `.nth()` on the iterator.

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.nth

http-api/src/lib.rs
sebastinez: According to `radicle_source::commits`, we just map the browser history to `vcs::git::Header`, did some tests and the committer times were not always sorted.
Eventually since it's almost sorted right, and to avoid resorting all commits we could avoid sorting and do it on the client side as currently.

http-api/src/lib.rs
sebastinez: Yeah you're right, will fix this

http-api/src/lib.rs
sebastinez: Ok nevermind you're right :+1:

http-api/src/lib.rs
sebastinez: Hmm ok strange, just checked again the committer times over a few repos and it looked correctly sorted, have removed the sorting now, to avoid sort repos with great amount of commits.

cloudhead (approved):
Nice.

PR #134, opened 2022-03-24
Pull --allow-unauthorized-keys option the .env file

adaszko:
The main use case is to do `--allow-unathorized-keys` for the community nodes.  Also remove unused GCP configuration.  Another thing is we limit maximum memory usage in git-server to prevent it taking down the entire cloud instance in case of it allocates unbounded amount of memory.  It's better if it gets killed by the kernel and the clients retry their requests.


.env
```diff
@@ -1,12 +1,2 @@
...
-RADICLE_INFLUXDB_TOKEN=ENC[AES256_GCM,data:19NKQrxIdCztW4mLh5/GfxZtAK4SBwYM2BRz/LSjFi2LPFYDuRITm/6R9DQ53va0aySG6ARM49AVcutOQz/Q94Sha21U2AmsHGvxs0cTYNDGMx+BQJcvww==,iv:XqIpHtSMXbjbD9EKUFDDkX2awpz4xFmCwL47Nx1+wVA=,tag:EEcg6E4CbmWAHyQK3bhRuw==,type:str]
-sops_gcp_kms__list_0__map_enc=CiQAw9HjBpHvRP7Kqe3v7vDzWKB1cRZGkGj78ZTwANxZG4iQX9USSQCYOiXnY7hYrqHYQnB51Y1kYph98xPFqD8gbwccE3OuVxkroLxd30FFTVUBcxxM+sykv4JDbv9VkkG5Dsm3zc+P4WhTsrxPl1s=
-sops_version=3.7.1
-sops_unencrypted_suffix=_unencrypted
-sops_gcp_kms__list_0__map_resource_id=projects/radicle-services/locations/global/keyRings/sops/cryptoKeys/sops-key
-sops_gcp_kms__list_0__map_created_at=2021-10-29T15:15:04Z
-sops_lastmodified=2021-10-29T15:15:04Z
-sops_mac=ENC[AES256_GCM,data:jD3xSgSAKzDg+HnM4h96/G51zB8YI3KPVCOENh2jiiwm9/m+p9+vcE8J9UXDO3mIEVfMy6WaniFgEn3g2eDqoxIDXij251rchzpaSNfwynUimDC9+3jibUpnD0fOe2Ky/1awd/dt/DC0kPaTG8u5cgM2FkuRSvxTGM0MF/wk5v0=,iv:KCvjnUo+qkFid2DHzpWy2fnXeVGq8vtr+VtiuoMMweY=,tag:p3aRTn06H6iIDB0yCfDYlg==,type:str]
+RADICLE_DOMAIN=
+GIT_SERVER_ENV_SPECIFIC_OPTS=
```
cloudhead, line 2: Can we call this `RADICLE_GIT_SERVER_OPTS`? I think it's clear that it's an env var already, but not that it's radicle-specific.

PR #141, opened 2022-04-11
http-api: Add sessions with SIWE authentication

sebastinez:
This PR adds new `sessions` route `handlers` and `filters` to allow users to sign into the `http-api`.

**Process:**
- We create a `Unauthorized` session for every POST request we receive on `/v1/sessions` that lasts for 60 seconds.
- Once we receive a PUT request with the SIWE raw message and the signature done by the user and verify the authenticity, origin and timestamps, we upgrade the session to a `Authorized` session and keep the session alive for 1 week
- We periodically check the active `Unauthorized` & `Authorized` sessions for expiration and remove them if necessary.

**Implementation:**
The idea is to provide the session id in the `Authorization` header, for the routes which we want to protect.

### DIFF
--- a/http-api/Cargo.toml
+++ b/http-api/Cargo.toml
@@ -15,6 +15,7 @@ serde_json = { version = "1", features = ["preserve_order"] }
 serde_urlencoded = { version = "0.7.0" }
 radicle-source = { version = "0.3.0", features = ["syntax"] }
 radicle-surf = { version = "0.7.0", features = ["serialize"] }
+siwe = "0.2"
 thiserror = { version = "1" }
 git2 = { version = "0.13", default-features = false, features = [] }
 tokio = { version = "1.2", features = ["macros", "rt", "sync"] }
@@ -23,6 +24,9 @@ either = { version = "1.6" }
 tracing = "0.1"
 tracing-subscriber = "0.2"
 async-trait = "0.1"
+ethers-core = "0.6.3"
+fastrand = "1.7.0"
+chrono = { version = "0.4.19", features = ["serde"] }
 
 [features]
 gcp = ["shared/gcp"]
--- a/http-api/src/auth.rs
+++ b/http-api/src/auth.rs
@@ -1,0 +1,59 @@
+use std::convert::TryFrom;
+
+use chrono::{DateTime, Utc};
+use ethers_core::types::{Signature, H160};
+use serde::{Deserialize, Serialize};
+
+use crate::error::Error;
+
+#[derive(Deserialize, Serialize)]
+pub struct AuthRequest {
+    pub message: String,
+    pub signature: Signature,
+}
+
+pub enum AuthState {
+    Authorized(Session),
+    Unauthorized {
+        nonce: String,
+        expiration_time: DateTime<Utc>,
+    },
+}
+
+// We copy the implementation of siwe::Message here to derive Serialization and Debug
+#[derive(Clone, Serialize)]
+pub struct Session {
+    pub domain: String,
+    pub address: H160,
+    pub statement: String,
+    pub uri: String,
+    pub version: u64,
+    pub chain_id: u64,
+    pub nonce: String,
+    pub issued_at: DateTime<Utc>,
+    pub expiration_time: Option<DateTime<Utc>>,
+    pub resources: Vec<String>,
+}
+
+impl TryFrom<siwe::Message> for Session {
+    type Error = Error;
+
+    fn try_from(message: siwe::Message) -> Result<Session, Error> {
+        let statement = message.statement.ok_or(Error::Auth("No statement found"))?;
+
+        Ok(Session {
+            domain: message.domain.host().to_string(),
+            address: H160(message.address),
+            statement,
+            uri: message.uri.to_string(),
+            version: message.version as u64,
+            chain_id: message.chain_id,
+            nonce: message.nonce,
+            issued_at: message.issued_at.as_ref().with_timezone(&Utc),
+            expiration_time: message
+                .expiration_time
+                .map(|x| x.as_ref().with_timezone(&Utc)),
+            resources: message.resources.iter().map(|r| r.to_string()).collect(),
+        })
+    }
+}
--- a/http-api/src/error.rs
+++ b/http-api/src/error.rs
@@ -28,29 +28,45 @@ pub enum Error {
     #[error("could not resolve head: {0}")]
     NoHead(&'static str),
 
-    /// An error occured with radicle identities.
+    /// An error occurred during an authentication process.
+    #[error("could not authenticate: {0}")]
+    Auth(&'static str),
+
+    /// An error occurred while verifying the siwe message.
+    #[error(transparent)]
+    SiweVerificationError(#[from] siwe::VerificationError),
+
+    /// An error occurred while parsing the siwe message.
+    #[error(transparent)]
+    SiweParseError(#[from] siwe::ParseError),
+
+    /// An error occurred with radicle identities.
     #[error(transparent)]
     Identities(#[from] librad::git::identities::Error),
 
-    /// An error occured with radicle surf.
+    /// An error occurred with radicle surf.
     #[error(transparent)]
     Surf(#[from] surf::git::error::Error),
 
-    /// An error occured with radicle storage.
+    /// An error occurred with radicle storage.
     #[error(transparent)]
     Storage(#[from] librad::git::storage::Error),
 
-    /// An error occured with radicle storage.
+    /// An error occurred with radicle storage.
     #[error("{0}: {1}")]
     Io(&'static str, std::io::Error),
 
-    /// An error occured with initializing read-only storage.
+    /// An error occurred with initializing read-only storage.
     #[error(transparent)]
     Init(#[from] librad::git::storage::read::error::Init),
 
-    /// An error occured with radicle source.
+    /// An error occurred with radicle source.
     #[error(transparent)]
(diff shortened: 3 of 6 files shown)


http-api/Cargo.toml
```diff
@@ -23,6 +24,8 @@ either = { version = "1.6" }
 tracing = "0.1"
 tracing-subscriber = "0.2"
 async-trait = "0.1"
+ethers-core = "0.6.3"
+rand = "0.8.5"
```
cloudhead, line 28: If you just need a simple rng, you can use `fastrand` which doesn't pull in a ton of dependencies.

http-api/src/auth.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct UnauthorizedMessage {
+    pub nonce: String,
+    pub expiration_time: u64,
+}
+
+impl From<siwe::Message> for AuthMessage {
+    fn from(message: siwe::Message) -> Self {
+        AuthMessage {
+            domain: message.domain.as_str().to_string(),
+            address: format!("{:?}", H160(message.address)),
```
cloudhead, line 41: It would be nice to not convert this to a string. Is the problem that it isn't `Serialize`?

http-api/src/auth.rs
```diff
@@ -0,0 +1,54 @@
...
+            statement: message.statement,
+            uri: message.uri.to_string(),
+            version: format!("{}", message.version as u64),
+            chain_id: message.chain_id.to_string(),
+            nonce: message.nonce.to_string(),
```
cloudhead, line 46: Why are these two converted to strings?

http-api/src/lib.rs
```diff
@@ -31,12 +39,18 @@ use librad::{git::types::Namespace, git::Urn, paths::Paths, profile::Profile, Pe
...
 
 use commit::{Commit, CommitContext, CommitTeaser, CommitsQueryString, Committer, Peer};
 use error::Error;
 
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
+pub const POPULATE_FINGERPRINTS_INTERVAL: u64 = 20;
+pub const CLEANUP_SESSIONS_INTERVAL: u64 = 60;
+pub const UNAUTHORIZED_SESSIONS_EXPIRATION: u64 = 60;
+// Authorized Sessions expire after 1 week
+pub const AUTHORIZED_SESSIONS_EXPIRATION: u64 = 604800;
```
cloudhead, line 53: Expirations should be of type `std::time::Duration`, you can use `Duration::from_secs` and `Duration::as_secs` to convert.

http-api/src/lib.rs
```diff
@@ -82,7 +97,65 @@ impl Context {
...
-    /// Populate a map between ssh fingerprints and their peer identities
+    fn cleanup_sessions(&self, map: &mut HashMap<String, AuthState>) -> Result<(), Error> {
+        let mut to_remove: Vec<String> = Vec::new();
+        let current_time = self.now()?;
+
+        for (key, value) in map.iter() {
+            match value {
+                AuthState::Authorized(auth) => {
+                    if let Some(exp_time) = auth.expiration_time {
+                        if exp_time < current_time {
```
cloudhead, line 108: Should be `>=` no?

http-api/src/lib.rs
```diff
@@ -82,7 +97,65 @@ impl Context {
...
+                            to_remove.push(key.clone());
+                        }
+                    }
+                }
+                AuthState::Unauthorized(auth) => {
+                    if auth.expiration_time < current_time {
```
cloudhead, line 114: Same here?

http-api/src/lib.rs
```diff
@@ -220,13 +321,115 @@ async fn peer_handler(peer_id: PeerId) -> Result<impl warp::Reply, warp::Rejecti
...
+    let session = sessions
+        .get(&id)
+        .ok_or_else(|| warp::reject::custom(Error::Siwe("Session not found")))?;
+
+    if let AuthState::Unauthorized(auth) = session {
+        let message = Message::from_str(request.message.as_str())
+            .map_err(|_| Error::Siwe("Not able to parse Siwe message"))?;
+        // TODO: Not sure if using an env variable is the right thing to do.
+        let host =
+            env::var("RADICLE_DOMAIN").map_err(|_| Error::Siwe("No host env variable found"))?;
```
cloudhead, line 373: This is probably fine for now, but I'd improve the error as `RADICLE_DOMAIN is not set`

cloudhead (commented):
Nice, did an initial review.

http-api/src/lib.rs
sebastinez: In this case we want to add expired sessions to a `to_remove` vector so we need to check if `exp_time` is smaller than the current timestamp.

http-api/src/lib.rs
sebastinez: In this case we are doing the same like earlier, I'll look into if we can avoid pushing session ids into an array over which we than have to iterate again, doing it a little bit more efficient.

http-api/src/auth.rs
sebastinez: Yeah should be possible to keep it a H160

http-api/src/lib.rs
```diff
@@ -406,6 +517,120 @@ async fn root_handler() -> Result<impl Reply, Rejection> {
     Ok(warp::reply::json(&response))
 }
 
+/// Return the peer id for the node identity.
+/// `GET /v1/peer`
+async fn peer_handler(peer_id: PeerId) -> Result<impl warp::Reply, warp::Rejection> {
```
sebastinez, line 522: Moved the peer handler down here to be together with the other handlers..

http-api/src/error.rs
```diff
@@ -28,29 +28,45 @@ pub enum Error {
...
     Init(#[from] librad::git::storage::read::error::Init),
 
-    /// An error occured with radicle source.
+    /// An error occurred with radicle source.
     #[error(transparent)]
     Source(#[from] radicle_source::error::Error),
+
+    /// An error occurred with env variables.
+    #[error(transparent)]
+    VarError(#[from] std::env::VarError),
```
sebastinez, line 69: This new Error is when `RADICLE_DOMAIN` is missing and is more explicit than a `warp::reject::custom` error IMO

http-api/src/error.rs
```diff
@@ -28,29 +28,45 @@ pub enum Error {
...
+    #[error("could not authenticate: {0}")]
+    Auth(&'static str),
+
+    /// An error occurred while verifying the siwe message.
+    #[error(transparent)]
+    SiweVerificationError(#[from] siwe::VerificationError),
+
+    /// An error occurred while parsing the siwe message.
+    #[error(transparent)]
+    SiweParseError(#[from] siwe::ParseError),
```
sebastinez, line 41: This seems maybe overkill but it's the only way I found to match against the `siwe` crate errors

http-api/src/lib.rs
```diff
@@ -211,22 +281,19 @@ async fn periodic(ctx: Context, interval: time::Duration) {
...
     } else if let Some(Error::NotFound) = err.find::<Error>() {
         (StatusCode::NOT_FOUND, None)
     } else if let Some(Error::NoHead(msg)) = err.find::<Error>() {
         (StatusCode::NOT_FOUND, Some(*msg))
+    } else if let Some(Error::Auth(msg)) = err.find::<Error>() {
+        (StatusCode::BAD_REQUEST, Some(*msg))
+    } else if let Some(Error::SiweParseError(_)) = err.find::<Error>() {
+        (StatusCode::BAD_REQUEST, None)
+    } else if let Some(Error::SiweVerificationError(_)) = err.find::<Error>() {
+        (StatusCode::BAD_REQUEST, None)
```
sebastinez, line 296: @cloudhead I was not entirely sure if we said to leave the Siwe Errors as `BAD_REQUEST` or to change some to `UNAUTHORIZED` IMO I think both are  the result of a bad request..

cloudhead (approved):
LGTM.

http-api/src/lib.rs
```diff
@@ -406,6 +511,92 @@ async fn root_handler() -> Result<impl Reply, Rejection> {
...
+
+    match session {
+        AuthState::Authorized(session) => {
+            Ok(warp::reply::json(&json!({ "id": id, "session": session })))
+        }
+        AuthState::Unauthorized {
+            nonce,
+            expiration_time,
+        } => Ok(warp::reply::json(
+            &json!({ "id": id, "nonce": nonce, "expiration_time": expiration_time }),
```
cloudhead, line 539: I think for json data we use camel case keys eg. `expirationTime`?

http-api/src/lib.rs
```diff
@@ -406,6 +511,92 @@ async fn root_handler() -> Result<impl Reply, Rejection> {
...
+    id: String,
+    request: AuthRequest,
+) -> Result<impl warp::Reply, warp::Rejection> {
+    // Get unauthenticated session data, return early if not found
+    let mut sessions = ctx.sessions.write().await;
+    let session = sessions.get(&id).ok_or(Error::NotFound)?;
+
+    if let AuthState::Unauthorized { nonce, .. } = session {
+        let message = Message::from_str(request.message.as_str())
+            .map_err(|_| Error::Auth("Not able to parse message"))?;
```
cloudhead, line 568: Is it a siwe::ParseError returned here?

sebastinez: Yes and we should be able to return a `Error::SiweParseError`
Good catch

PR #144, opened 2022-04-26
http-api: List patches

### LINKED ISSUE #143: http-api: List patches
We could make use of a new endpoint on the http-api that would allow us to retrieve the open and closed patches.
This would allow us to show a listing of patches on the web client similar to upstream, and eventually even link to upstream.

I propose to add the following route:
`GET /v1/patches/<project-urn>/patches // returns all patches from a specific project`

sebastinez:
This PR adds a new filter and handler to able to query all the patches stored from the tracked peers.

Closes #143

### DIFF
--- a/http-api/src/lib.rs
+++ b/http-api/src/lib.rs
@@ -461,6 +461,14 @@ async fn patches_handler(ctx: Context, urn: Urn) -> Result<impl Reply, Rejection
             for blob in blobs {
                 if let Some(object) = storage.find_object(blob.1).map_err(Error::from)? {
                     let tag = object.peel_to_tag().map_err(Error::from)?;
+                    // Only match tags that have a valid utf8 name that starts with patches/
+                    if !tag
+                        .name()
+                        .ok_or(Error::Utf8Error)?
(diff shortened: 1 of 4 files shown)


http-api/src/lib.rs
```diff
@@ -427,6 +443,51 @@ async fn blob_handler(
...
+            };
+            // Iterate over all refs, and return eventual tags aka patches
+            for blob in blobs {
+                if let Some(object) = storage.find_object(blob.1).map_err(Error::from)? {
+                    let tag = object.peel_to_tag().map_err(Error::from)?;
+                    // Only match tags that have a valid ut8 name that starts with patches/
+                    if !tag
+                        .name()
+                        .ok_or(Error::Utf8Error)?
+                        .starts_with("radicle-patch/")
```
rudolfs, line 468: This doesn't match what the comment says. 🤔

sebastinez: Thanks, slipped through the cracks!

sebastinez: Thanks, slipped through the cracks!

PR #148, opened 2022-05-11
Update radicle-common patch rev

sebastinez:
Updates the revision used in the radicle-common patch section.
Without it the http-api crashes when trying to read comments.


cloudhead:
Superseded by 815398

PR #154, opened 2022-06-02
http-api: Resolve issue and comment authors

sebastinez:
Signed-off-by: Sebastian Martinez <me@sebastinez.dev>

### DIFF
--- a/http-api/src/error.rs
+++ b/http-api/src/error.rs
@@ -38,11 +38,11 @@ pub enum Error {
 
     /// An error occurred while verifying the siwe message.
     #[error(transparent)]
-    SiweVerificationError(#[from] siwe::VerificationError),
+    SiweVerification(#[from] siwe::VerificationError),
 
     /// An error occurred while parsing the siwe message.
     #[error(transparent)]
-    SiweParseError(#[from] siwe::ParseError),
+    SiweParse(#[from] siwe::ParseError),
 
     /// An error occurred with radicle identities.
     #[error(transparent)]
@@ -86,11 +86,11 @@ pub enum Error {
 
     /// An error occurred with env variables.
     #[error(transparent)]
-    VarError(#[from] std::env::VarError),
+    Env(#[from] std::env::VarError),
 
     /// An error occurred during the identity resolving.
     #[error(transparent)]
-    IdentityResolveError(#[from] radicle_common::cobs::ResolveError),
+    IdentityResolve(#[from] radicle_common::cobs::ResolveError),
 }
 
 impl warp::reject::Reject for Error {}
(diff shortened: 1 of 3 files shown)


http-api/src/issues.rs
```diff
@@ -68,10 +77,13 @@ async fn issue_handler(
...
     let whoami = person::local(&*storage).map_err(Error::LocalIdentity)?;
     let issues = issue::Issues::new(whoami, &ctx.paths, &storage).map_err(Error::Issues)?;
-    let issue = issues
+    let mut issue = issues
         .get(&project, &issue_id)
         .map_err(Error::Issues)?
         .ok_or(Error::NotFound)?;
+    issue
+        .resolve(storage.as_ref())
+        .map_err(Error::IdentityResolveError)?;
```
cloudhead, line 86: Do we want to bail here? I'm not sure. We should still be able to render the issue without a resolved author I would think?

http-api/src/issues.rs
```diff
@@ -53,7 +53,16 @@ async fn issues_handler(ctx: Context, project: Urn) -> Result<impl Reply, Reject
         .all(&project)
         .map_err(Error::Issues)?
         .into_iter()
-        .map(|(id, issue)| Cob::new(id, issue))
+        .map(|(id, mut issue)| {
+            if let Err(e) = issue
+                .resolve(storage.as_ref())
+                .map_err(Error::IdentityResolveError)
+            {
+                tracing::debug!("Failed to resolve issue author: {}", e);
```
cloudhead, line 61:
```suggestion
                tracing::error!("Failed to resolve issue author: {}", e);
```

cloudhead: Would be good for this to stick out in the logs, or maybe `warn` is enough.

cloudhead: Also might be handy to know the issue id in the log.

http-api/src/error.rs
```diff
@@ -87,6 +87,10 @@ pub enum Error {
     /// An error occurred with env variables.
     #[error(transparent)]
     VarError(#[from] std::env::VarError),
+
+    /// An error occurred during the identity resolving.
+    #[error(transparent)]
+    IdentityResolveError(#[from] radicle_common::cobs::ResolveError),
```
cloudhead, line 93: We usually avoid repeating `Error` in the variants, so I'd rename this to `IdentityResolve`. You could also fix the one above to just `Var` or `Env` maybe.

http-api/src/issues.rs
sebastinez: No you're right, we should just warn about it, for debugging purposes

http-api/src/issues.rs
sebastinez: Great, yeah that makes sense.

http-api/src/error.rs
sebastinez: Ok great, I also fixed the `SiweVerificationError` and the `SiweParseError`

PR #162, opened 2022-06-21
Various patches endpoint updates

### LINKED ISSUE #155: Create endpoints for issue and patch count
We should add the issue and patches count to the project `Metadata`.
This will allows us to query the patches and issue count, without fetching all patches or issues.

To have a more complete picture we could also return an object with the following structure:

```json
{
  "issues": {
    "open": 9,
    "closed": {
        "solved": 3,
        "other": 4
    }
  }
}
```

sebastinez:
This PR closes #155 

- Converts patch endpoints from using tags to COBs.
- Adds an endpoint for single patch retrieval
- Adds partial and complete diffs

### DIFF
--- a/http-api/src/patches.rs
+++ b/http-api/src/patches.rs
@@ -1,12 +1,21 @@
+use std::convert::TryFrom;
+use std::convert::TryInto;
+
 use warp::{self, path, Filter, Rejection, Reply};
 
 use librad::collaborative_objects::ObjectId;
-use librad::git::Urn;
+use librad::git::{identities, Storage, Urn};
+use librad::PeerId;
 
 use radicle_common::cobs::{patch, Store};
-use radicle_common::person;
+use radicle_common::{person, project};
+use radicle_source::commit::Stats;
+use radicle_source::surf::vcs::git;
+use radicle_source::Commit;
+use radicle_surf::diff;
 
 use crate::error::Error;
+use crate::remote_branch;
 use crate::Context;
 
 /// A collaborative object that includes its id.
@@ -23,6 +32,72 @@ impl<T: serde::Serialize> Cob<T> {
     }
 }
 
+#[derive(serde::Serialize, Clone)]
+struct Changeset {
+    commits: Vec<Commit>,
+    diff: git::Diff,
+    stats: Stats,
+}
+
+impl Changeset {
+    pub fn new(commits: Vec<Commit>, diff: git::Diff) -> Self {
+        Self {
+            commits,
+            stats: Changeset::stats(&diff),
+            diff,
+        }
+    }
+
+    // TODO: This function should probably be moved to radicle_surf, where it should be able to be called on radicle_surf::diff::Diff as associated function`
+    pub fn stats(diff: &git::Diff) -> Stats {
+        let mut deletions = 0;
+        let mut additions = 0;
+
+        for file in &diff.modified {
+            if let diff::FileDiff::Plain { ref hunks } = file.diff {
+                for hunk in hunks.iter() {
+                    for line in &hunk.lines {
+                        match line {
+                            diff::LineDiff::Addition { .. } => additions += 1,
+                            diff::LineDiff::Deletion { .. } => deletions += 1,
+                            _ => {}
+                        }
+                    }
+                }
+            }
+        }
+
+        for file in &diff.created {
+            if let diff::FileDiff::Plain { ref hunks } = file.diff {
+                for hunk in hunks.iter() {
+                    for line in &hunk.lines {
+                        if let diff::LineDiff::Addition { .. } = line {
+                            additions += 1
+                        }
+                    }
+                }
+            }
+        }
+
+        for file in &diff.deleted {
+            if let diff::FileDiff::Plain { ref hunks } = file.diff {
+                for hunk in hunks.iter() {
+                    for line in &hunk.lines {
+                        if let diff::LineDiff::Deletion { .. } = line {
+                            deletions += 1
+                        }
+                    }
+                }
+            }
+        }
+
+        Stats {
+            additions,
+            deletions,
+        }
+    }
+}
+
 /// `GET /:project/patches`
 pub fn patches_filter(
     ctx: Context,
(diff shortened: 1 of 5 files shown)


http-api/src/patches.rs
```diff
@@ -23,6 +33,32 @@ impl<T: serde::Serialize> Cob<T> {
     }
 }
 
+/// A wrapper struct to include diffs in revisions
+#[derive(serde::Serialize)]
+struct RevisionWithDiffs {
```
cloudhead, line 38: `RevsiionWithDiff` (singular)

http-api/src/patches.rs
```diff
@@ -23,6 +33,32 @@ impl<T: serde::Serialize> Cob<T> {
...
+    diff: diff::Diff,
+    merges: Vec<MergeWithPeerInfo>,
+}
+
+/// A merged patch revision.
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct MergeWithPeerInfo {
+    #[serde(flatten)]
+    pub merge: Merge,
+    pub peer: PeerId,
```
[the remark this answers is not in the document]
sebastinez: Yes this has been fixed in the latest changes

http-api/src/patches.rs
```diff
@@ -20,19 +55,147 @@ pub fn patches_filter(
...
+                discussion: revision.discussion.clone(),
+                reviews: revision.reviews.clone(),
+                timestamp: revision.timestamp,
+                changeset: Changeset::new(commits, diff),
+                merges,
+            }
+        })
+        .collect::<Vec<_>>()
+        .try_into()
+        .unwrap(); // This unwrap is save, since we work with a NonEmpty struct and won't collect an empty Vec.
```
cloudhead, line 188:
```suggestion
        .unwrap(); // This unwrap is safe, since we work with a NonEmpty struct and won't collect an empty Vec.
```

cloudhead:
You can rebase this now, I merged the patch count pr.

http-api/src/patches.rs
```diff
@@ -52,9 +100,145 @@ async fn patches_handler(ctx: Context, urn: Urn) -> Result<impl Reply, Rejection
...
+
+    let patch = patch::Patch {
+        author: patch.author,
+        title: patch.title,
+        state: patch.state,
+        target: patch.target,
+        labels: patch.labels,
+        timestamp: patch.timestamp,
+        revisions,
+    };
```
cloudhead, line 168: Since you're moving out of `Patch`, I would pass it into `resolve_revisions` by value, and return it. That way you don't need to clone all the fields like you do bellow.

cloudhead: I think you would have to use `into_iter` as well

http-api/src/patches.rs
```diff
@@ -52,9 +100,145 @@ async fn patches_handler(ctx: Context, urn: Urn) -> Result<impl Reply, Rejection
                 tracing::warn!("Failed to resolve identities in patch {}: {}", id, e);
             }
 
+            let revisions = resolve_revisions(&patch, &mut browser, &meta, storage.as_ref());
```
cloudhead, line 103: Why do we need the changeset when listing patches?

http-api/src/patches.rs
```diff
@@ -23,6 +35,18 @@ impl<T: serde::Serialize> Cob<T> {
     }
 }
 
+#[derive(serde::Serialize, Clone)]
+struct Changeset {
+    commits: Option<Vec<Commit>>,
+    diff: Option<git::Diff>,
+}
```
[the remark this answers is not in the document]
sebastinez: Sounds right, in this case though if the total diff fails the Changeset option  would need to be `None` and vice versa.
But eventually we wouldn't want partial commit diffs or the entire diff either way if one of the operations fail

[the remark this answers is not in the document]
cloudhead: Exactly.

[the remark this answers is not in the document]
cloudhead: It's also unlikely that one fails and not the other, because they are accessing the same objects mostly.

http-api/src/patches.rs
cloudhead: Yes, but you avoid cloning the entire discussion :)

sebastinez:
@cloudhead Please let me know if you have any other observations, I feel like we are at a good place, with just enough room for improvement hehe
Will do a rebase + squash when you are okay with it

http-api/src/patches.rs
```diff
@@ -52,9 +160,149 @@ async fn patches_handler(ctx: Context, urn: Urn) -> Result<impl Reply, Rejection
...
+        .into_iter()
+        .map(|revision| {
+            // Locates the browser at the Oid of the current revision.
+            if let Err(e) = browser.rev(Rev::Oid(*revision.oid)) {
+                tracing::warn!(
+                    "Failed to set the browser history at {}: {}",
+                    *revision.oid,
+                    e
+                );
+            }
```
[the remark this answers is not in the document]
sebastinez: Hmm we should eventually resolve the merges first and then if there are issues with looking up the revision's Oid return the patch with no changeset.

sebastinez:
Ok did a squash and rebase and fixed the last issues, and found some things to clean up make tidier

http-api/src/patches.rs
```diff
@@ -52,9 +156,162 @@ async fn patches_handler(ctx: Context, urn: Urn) -> Result<impl Reply, Rejection
...
+                            .filter_map(|header| radicle_source::commit(browser, header.sha1).ok())
+                            .collect::<Vec<Commit>>(),
+                    )
+                } else {
+                    None
+                };
+
+                // Gets the entire diff between the default branch head and the revision Oid.
+                // TODO: Probably we can handle a possible error better, but for now we handle it in the interface
+                let diff = browser.diff(*revision.base, *revision.oid).ok();
```
cloudhead, line 286: There's no need to do a diff if `commits == None`, since the changeset will be `None`! Sorry didn't notice this before :)

cloudhead: I would probably simplify `Changeset::new` and just have it accept the types without the `Option` as well, otherwise we're allowing stuff like this.

sebastinez: Nice, good catch, was able to reduce nesting by removing if branches and make the code clearer in 36522db5b0a90a8216ecb3f222f8b2a3a7fac9c8

PR #164, opened 2022-06-27
Fixes related to server-side profiles

adaszko:


git-server/Dockerfile
```diff
@@ -9,18 +9,19 @@ COPY . .
...
 
-WORKDIR /usr/src/radicle-client-services/service-init
-RUN cargo install --profile=container --all-features --locked --path .
-
 # Run
 FROM debian:bullseye-slim@sha256:4c25ffa6ef572cf0d57da8c634769a08ae94529f7de5be5587ec8ce7b9b50f9c
 
+RUN apt update && apt install --yes curl gnupg2 && rm -rf /var/lib/apt/lists/*
+RUN curl https://europe-west6-apt.pkg.dev/doc/repo-signing-key.gpg | apt-key add -
+RUN echo deb https://europe-west6-apt.pkg.dev/projects/radicle-services radicle-cli main | tee -a /etc/apt/sources.list.d/radicle-registry.list
```
cloudhead, line 17: Are we installing the CLI just so we can run `rad-self --profile`? We can avoid that actually if we do `$(cat $RAD_HOME/active_profile)`

cloudhead (approved):

PR #168, opened 2022-07-11
axum: adding routes

xphoniex:


xphoniex:
Current arrangement cleans up `lib.rs` as it moves relevant fns to where they should be. E.g. if and when we want to add `/v2/` we simply add `.merge(v2::router(ctx.clone()))` to our `app` in lib.rs and leave the rest of it alone.

Each route can be easily found, as you need to look in their respective file. In total we'll create 5 more files but it'd be worth it as helper functions will also move to where they belong, and so will imports.

This will make each file more specialized and more readable.

http-api/src/v1/peer.rs
```diff
@@ -0,0 +1,23 @@
...
+pub fn router(ctx: Context) -> Router {
+    let peer_id = ctx.peer_id;
+
+    Router::new()
+        .route("/peer", get(peer_handler))
+        .layer(Extension(peer_id))
+}
+
+/// Return the peer id for the node identity.
+/// `GET /peer`
```
xphoniex, line 16: Currently some routes are commented with prefix e.g. `GET /v1/peer` and some are not. I'll rewrite all with no prefix but lmk if you want otherwise. (Full path should be obvious, by looking at the file path)

http-api/src/v1/sessions.rs
```diff
@@ -0,0 +1,177 @@
...
+    Router::new()
+        .route("/sessions", post(session_create_handler))
+        .route(
+            "/sessions/:session_id",
+            get(session_get_handler).put(session_signin_handler),
+        )
+        .layer(Extension(ctx.clone()))
+}
+
+/// TODO: Add description.
```
xphoniex, line 79: Some handlers are missing description. We want the same format above each handler, right?

http-api/src/v1/sessions.rs
```diff
@@ -0,0 +1,177 @@
...
+        Utc::now() + chrono::Duration::from_std(UNAUTHORIZED_SESSIONS_EXPIRATION).unwrap();
+    let mut sessions = ctx.sessions.write().await;
+    let (session_id, nonce) = create_session(&mut sessions, expiration_time);
+    let response = Json(json!({ "id": session_id, "nonce": nonce }));
+
+    response
+}
+
+/// TODO: Add description.
+/// `GET /sessions/:session_id`
```
xphoniex, line 92: any reason why we're using a `:session_id` and not an `:id`? we only have one in this path, why name it?

http-api/src/v1/sessions.rs
xphoniex: Also what should the description be for:
- `GET /sessions/:session_id`
- `PUT /sessions/:session_id`
- `POST /sessions`

http-api/src/v1/peer.rs
xphoniex: Also [project routes](https://github.com/radicle-dev/radicle-client-services/blob/29416ea9dc207074b4d9c84648fc648f46ce0bf2/http-api/src/lib.rs#L441) lack the `/project` prefix which I'll add.

xphoniex:
Need descriptions for:

- `GET /projects/:project/commits/:sha`
- `GET /projects/:project/commits?from=<sha>`
- `GET /projects/:project-urn or :project-alias`
- `GET /projects/:project/tree/:prefix/*path`
- `GET /projects/:project/remotes`
- `GET /projects/:project/remotes/:peer`
- `GET /projects/:project/blob/:sha/*path?highlight=<bool>`
- `GET /projects/:project/readme/:sha`

http-api/src/v1/sessions.rs
cloudhead: Get session, update session, create session.

http-api/src/v1/sessions.rs
cloudhead: Should be `:id` yeah.

http-api/src/custom_axum.rs
```diff
@@ -0,0 +1,57 @@
+use axum::extract::path::ErrorKind;
```
cloudhead, line 1: If this is an "axum related" file, maybe we can just call it `axum.rs`

cloudhead:
LGTM!

cloudhead (approved):

http-api/src/custom_axum.rs
xphoniex: Then I have to use `mod axum` in lib.rs which interferes with the extern crate of `axum` itself.

For now I changed filename to axum.rs and import it with:
```
#[path = "axum.rs"]
mod custom_axum;
```

cloudhead: Ah gotcha yes. I guess by convention these are called "extra" modules, eg. we have a `git_extra` module in radicle link, so maybe `axum_extra` would make sense here?

http-api/src/v1/sessions.rs
```diff
@@ -0,0 +1,171 @@
...
+            nonce,
+            expiration_time,
+        } => Ok::<_, Error>(Json(
+            json!({ "id": id, "nonce": nonce, "expirationTime": expiration_time }),
+        )),
+    }
+}
+
+/// Update session.
+/// `PUT /sessions/:session_id`
```
cloudhead, line 110: Don't forget to change the `:session_id`in the docstrings as well

http-api/src/v1/projects.rs
```diff
@@ -488,14 +488,21 @@ async fn remote_handler(
     Ok::<_, Error>(Json(response))
 }
 
+#[derive(Deserialize, Default)]
+struct HighlightQuery {
```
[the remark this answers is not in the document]
xphoniex: https://github.com/radicle-dev/radicle-client-services/pull/168/commits/adeb5cccfbf2a8bf4d496e26603e2edfbc920105

http-api/src/v1/sessions.rs
xphoniex: https://github.com/radicle-dev/radicle-client-services/pull/168/commits/a3779ffffb7f1e3d6044a5ac0628bd359def3b69

PR #170, opened 2022-07-25
http-api: Add comment route

sebastinez:
**Pending:** 
- [ ] Return Error when patch or issue is not found.
- [ ] Improve the return value (atm nothing is returned)


sebastinez:
Will close this PR for now, until we have a more clear sight on how we want to handle the authentication between the seed node and the user.

PR #174, opened 2022-08-12
http-api: Add project activity endpoint

sebastinez:
This PR adds an endpoint at `/v1/projects/:project/activity` that returns the committer timestamps of all commits done in the past year, starting at the project HEAD.

Since we are not mapping `Commit` to `Header` or doing other filtering and sorting, this should be a fast way to obtain the repo activity.

The grouping by week, I left to be done by `radicle-interface` since we would eventually be interested in grouping or sorting these timestamps differently.


cloudhead (approved):
Looks good.

PR #177, opened 2022-08-13
http-api: Add `STATUS` to response logging

xphoniex:

PR #178, opened 2022-08-18
git-server: migrate from `warp` to `axum`

xphoniex:


git-server/src/error.rs
```diff
@@ -146,4 +155,10 @@ impl Error {
     }
 }
 
-impl warp::reject::Reject for Error {}
+impl IntoResponse for Error {
```
xphoniex, line 158: I didn't use `recover` implementation since `self` is an `Error` and has no `is_not_found()` method. This `IntoResponse` only returns status codes defined above in `pub fn status(&self)`.

git-server/src/lib.rs
```diff
@@ -305,54 +318,68 @@ pub async fn run(options: Options) -> anyhow::Result<()> {
...
-        .and(path::peek())
-        .and(
-            warp::filters::query::raw()
-                .or(warp::any().map(String::default))
-                .unify(),
+    let app = git_handler
+        .layer(Extension(ctx.clone()))
+        .layer(
+            TraceLayer::new_for_http()
+                .make_span_with(DefaultMakeSpan::new().level(Level::INFO))
```
xphoniex, line 325: I opted for response only logging, with default level set to `INFO`. Logging now looks like:

```
Aug 18 19:45:05.308  INFO radicle_git_server: version 0.1.0-7c60de6
Aug 18 19:45:05.309  INFO radicle_git_server: git version 2.34.1
Aug 18 19:45:05.309  INFO shared: Profile 7a84311f-a61b-4700-8e3f-0ff3fe65eeee loaded...
Aug 18 19:45:05.363  INFO radicle_git_server: alias "nakamoto" for "rad:git:hnrkgswoskaqzdc1b3f8trncgaw736nfbjxdy"
Aug 18 19:45:05.365  INFO radicle_git_server: listening on http://0.0.0.0:8778
Aug 18 19:45:08.042  INFO request{method=GET uri=/nakamoto.git/info/refs?service=git-upload-pack version=HTTP/1.1}: radicle_git_server: git-http-backend: exited successfully for rad:git:hnrkgswoskaqzdc1b3f8trncgaw736nfbjxdy
Aug 18 19:45:08.043  INFO request{method=GET uri=/nakamoto.git/info/refs?service=git-upload-pack version=HTTP/1.1}: radicle_git_server: STATUS=200 LATENCY=11.613317ms
Aug 18 19:45:08.060  INFO request{method=POST uri=/nakamoto.git/git-upload-pack version=HTTP/1.1}: radicle_git_server: git-http-backend: exited successfully for rad:git:hnrkgswoskaqzdc1b3f8trncgaw736nfbjxdy
Aug 18 19:45:08.061  INFO request{method=POST uri=/nakamoto.git/git-upload-pack version=HTTP/1.1}: radicle_git_server: STATUS=200 LATENCY=13.540688ms
```

A bit long, but now there's only one line per hit. The other lines are warnings/logs/etc.

xphoniex:
Both `git clone` and `rad sync` work with the new implementation. However I was not able to test any request which used `PeerId` in its `Uri` to check the correctness.

In what case is `PeerId` used?

git-server/src/lib.rs
```diff
@@ -305,54 +318,68 @@ pub async fn run(options: Options) -> anyhow::Result<()> {
...
+    uri: Uri,
+    query: RawQuery,
+) -> impl IntoResponse {
+    let query = query.0.unwrap_or_else(|| "".to_string());
+    let (peer_id, project_id, request) = {
+        let tokens = uri.path().split('/').collect::<Vec<&str>>();
+        let peer_id: Option<PeerId> = tokens.get(0).map(|p| p.parse().ok()).unwrap();
+        let (project_id, request) = if peer_id.is_none() {
+            if tokens.len() < 3 {
+                return Err(Error::ServiceUnavailable(""));
```
xphoniex, line 368: Also, is empty error msg here fine?

git-server/src/lib.rs
```diff
@@ -305,54 +318,68 @@ pub async fn run(options: Options) -> anyhow::Result<()> {
...
+            .serve(app)
+            .await
+            .unwrap();
     } else {
-        server.run(options.listen).await;
+        tracing::info!("listening on http://{}", options.listen);
+        axum::Server::bind(&options.listen)
+            .serve(app)
+            .await
+            .unwrap();
```
cloudhead, line 347: Can we return errors instead of unwrapping in this func?

git-server/src/lib.rs
```diff
@@ -305,54 +318,68 @@ pub async fn run(options: Options) -> anyhow::Result<()> {
...
+        } else {
+            if tokens.len() < 4 {
+                return Err(Error::ServiceUnavailable(""));
+            }
+
+            (tokens[2], tokens[3..].join("/"))
+        };
+
+        (peer_id, project_id, request)
+    };
```
cloudhead, line 381: Not a fan of manually parsing the URL path... there must be a better way, or perhaps for now drop support for optional PeerId, as we aren't using it yet.

git-server/src/lib.rs
```diff
@@ -305,54 +318,68 @@ pub async fn run(options: Options) -> anyhow::Result<()> {
...
-    request: warp::filters::path::Peek,
-    query: String,
-) -> Result<Box<dyn Reply>, Rejection> {
-    let remote = remote.expect("There is always a remote for HTTP connections");
+    body: Bytes,
+    ConnectInfo(remote): ConnectInfo<SocketAddr>,
+    uri: Uri,
+    query: RawQuery,
+) -> impl IntoResponse {
+    let query = query.0.unwrap_or_else(|| "".to_string());
```
cloudhead, line 362: `unwrap_or_default()`?

git-server/src/lib.rs
xphoniex: @cloudhead do you prefer it this way? Cause I have to change https://github.com/radicle-dev/radicle-client-services/pull/177 as well.

cloudhead: Responded in the other PR, I'm not sure what "response only logging" means..

git-server/src/lib.rs
xphoniex: `PeerId` is not the issue here per se. I'm using [`MakeService`](https://github.com/radicle-dev/radicle-client-services/blob/221ab2a75a7b7774964a64eded3f510cca480713/git-server/src/lib.rs#L332) (which was suggested in axum channel) since we don't need a `Router`.

If you find this problematic, I can use `any` in combination with `Router` but that has performance cost.

Personally, I'd rather write tests for extraction logic than take performance hit, but I can rewrite it to take `Router`/`any` if you want.

git-server/src/lib.rs
xphoniex: https://github.com/radicle-dev/radicle-client-services/compare/221ab2a75a7b7774964a64eded3f510cca480713..db63e7a4f28c0d0e7beb5b1a1cf2339d6129722d#diff-b672a15ad14f54528cebd77aee76d3007962d44bda987b3e0a34597ec80fd6b1R352-R355

git-server/src/lib.rs
xphoniex: https://github.com/radicle-dev/radicle-client-services/compare/221ab2a75a7b7774964a64eded3f510cca480713..db63e7a4f28c0d0e7beb5b1a1cf2339d6129722d#diff-b672a15ad14f54528cebd77aee76d3007962d44bda987b3e0a34597ec80fd6b1R370

git-server/src/lib.rs
cloudhead: IMO it's the job of the router/framework to parse URLs, not us. I'd either drop the peer-id parsing or take the performance hit (which I doubt is even measurable).

xphoniex: https://github.com/radicle-dev/radicle-client-services/compare/479c88c6c5672ee135534cf98191b991e0d12c77..cd2a351c51c38d58db8473998e0aefdfe1f168db

xphoniex: `PeerId` logic is in there, so we can add that back in, in future.

cloudhead (approved):

PR #179, opened 2022-08-19
Fix for: random requests pause for multiple seconds

adaszko:


http-api/src/lib.rs
```diff
@@ -289,15 +288,21 @@ async fn cleanup_sessions_job(ctx: Context, interval: time::Duration) {
...
     loop {
         timer.tick().await; // Returns immediately the first time.
 
-        let mut projects = ctx.projects.write().await;
-        if let Err(err) = ctx.populate_fingerprints(&mut projects).await {
-            tracing::error!("Failed to populate project fingerprints: {}", err);
+        let mut projects_guard = ctx.projects.write().await;
+        let projects = projects_guard.clone();
+        let storage = ctx.storage().await?;
+        *projects_guard = match tokio::task::spawn_blocking(move || Context::populate_fingerprints(projects, storage)).await.unwrap() {
```
adaszko, line 300: This `.unwrap()` needs to go of course.

cloudhead:
Can we use `std::Mutex` to avoid the cloning?

adaszko:
> Can we use `std::Mutex` to avoid the cloning?

No, I don't think that's possible.  `spawn_blocking()` requires its closure argument to be `'static` and the temporary lock guard created via `ctx.projects.write().await` is of a smaller lifetime than `'static'`.

What I've done instead to avoid copying is I've replaced the call to `spawn_blocking()` with `block_in_place()`.   It fixes the original problem and avoids copying.

PR #182, opened 2022-08-25
http-api: Add project list pagination

sebastinez:
This PR adds pagination to the project root handler.

- Creates chunks of size defined by query param `per_page` and defaults to 10 projects per page.
- Gets chunk at index defined by query param `page` and defaults to first page if None given.

We had some use cases where we queried for all projects of a seed, but currently we removed all this use cases.
If this still would be necessary, I would look for a way.

Also adds a project total count property to a new `/stats` route.

### DIFF
--- a/http-api/src/v1/stats.rs
+++ b/http-api/src/v1/stats.rs
@@ -1,0 +1,31 @@
+use axum::response::IntoResponse;
+use axum::routing::get;
+use axum::{Extension, Json, Router};
+use librad::git::identities::{self, SomeIdentity};
+use serde_json::json;
+
+use crate::Context;
+use crate::Error;
+
+pub fn router(ctx: Context) -> Router {
+    Router::new()
+        .route("/stats", get(stats_handler))
+        .layer(Extension(ctx))
+}
+
+/// Return the stats for the node.
+/// `GET /stats`
+async fn stats_handler(Extension(ctx): Extension<Context>) -> impl IntoResponse {
+    let storage = ctx.storage().await?;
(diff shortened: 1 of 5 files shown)


http-api/src/v1/stats.rs
```diff
@@ -0,0 +1,26 @@
...
+/// Return the stats for the node.
+/// `GET /stats`
+async fn stats_handler(Extension(ctx): Extension<Context>) -> impl IntoResponse {
+    let storage = ctx.storage().await?;
+    let project_count = identities::any::list(storage.read_only())
+        .map_err(Error::from)?
+        .filter(|res| matches!(res, Ok(SomeIdentity::Project(_))))
+        .count();
+
+    Ok::<_, Error>(Json(json!({ "project_count": project_count })))
```
cloudhead, line 25: Perhaps we can create some hierarchy here and have `{ project: { count: 123 } }`, this creates the space for other stats in the future.

sebastinez: Yeah for sure, wanted to add also the following we talked about
> we can have { projects: { count: 123 }, users: { count: 432 } }

sebastinez: Done in [528f84e](https://github.com/radicle-dev/radicle-client-services/pull/182/commits/528f84eb6588d75324acc19caf9abaa30ca6b160)

cloudhead:
Looks good, do you want to squash this to 1 or 2 commits?

sebastinez:
> Looks good, do you want to squash this to 1 or 2 commits?

Yeah will do that 👍

PR #183, opened 2022-08-26
http-api: allow specifying color theme in http request

rudolfs:
We allow setting the theme via the request, instead of configuring it via the CLI.
This allows us to implement theme switching in the UI, see: https://github.com/radicle-dev/radicle-interface/pull/298#issuecomment-1225715225.

The request would look like this:
http://127.0.0.1:8777/v1/projects/rad:git:hnrk8ueib11sen1g9n1xbt71qdns9n4gipw1o/blob/8604dc616b8531c672f4e1358a00770957bfbe94/svelte.config.js?highlight=true&theme=base16-ocean.light

If the `theme` query param isn't passed in, we fall back to the default `base16-ocean.dark` theme.

### DIFF
--- a/http-api/src/v1/projects.rs
+++ b/http-api/src/v1/projects.rs
@@ -414,6 +414,7 @@ async fn remote_handler(
 #[derive(Deserialize, Default)]
 struct BlobQuery {
     highlight: bool,
+    theme: Option<String>,
 }
 
 /// Get project source file.
@@ -426,7 +427,7 @@ async fn blob_handler(
     let path = path.strip_prefix('/').ok_or(Error::NotFound)?.to_string();
     let Query(query) = query.unwrap_or_default();
     let theme = if query.highlight {
-        Some(ctx.theme.as_str())
+        Some(query.theme.as_deref().unwrap_or(ctx.theme.as_str()))
     } else {
         None
     };


http-api/src/v1/projects.rs
```diff
@@ -380,7 +381,7 @@ async fn blob_handler(
     let path = path.strip_prefix('/').ok_or(Error::NotFound)?.to_string();
     let Query(query) = query.unwrap_or_default();
     let theme = if query.highlight {
-        Some(ctx.theme.as_str())
+        Some(query.theme.as_deref().unwrap_or("base16-ocean.dark"))
```
cloudhead, line 384: I like this idea, but I would leave the theme in the context as the default/fallback. That way it can still be configured when nothing's passed in the request.

rudolfs: What would be the use case? There has to be some kind of front-end through which we'd look at the code anyway and then the frontend should always be able to control what theme to show, no?

cloudhead: Well the frontend doesn't necessarily know the available themes right? There has to be a way to just ask for the default theme, which currently you do by not setting `theme=`, but then the question is who sets the default theme?

rudolfs: By default the API responds without including any colors. You have to explicitly set `?highlight=true`, and in that case I assume the frontend knows about color themes and would pass in the theme that it requires. You're right in that we don't have an endpoint listing available themes, but since `radicle-interface` is the only consumer at the moment, we can just assume the frontend knows what themes are available. Wouldn't falling back to the dark theme as we do now be enough?

If you really insist, I can add back the CLI param. I just don't fully understand why a node operator would ever want to set the theme color. 🤷

cloudhead: > If you really insist, I can add back the CLI param. I just don't fully understand why a node operator would ever want to set the theme color, when the frontend will override it anyway. shrug

Because you made it optional, so the frontend may not override it.. Either make it required, and then we don't need a default, or make it optional, and then we need a way to set the default. Also, if it's non-optional, you may as well collapse the two into a single var, eg. `highlight=base16-ocean.dark`

rudolfs: > Also, if it's non-optional, you may as well collapse the two into a single var, eg. highlight=base16-ocean.dark

Yeah, I thought about that, but I didn't want to break the existing UI if we merge and deploy this before the theme PR is merged on interface.

I still think that we don't need to be able to set the default theme via a CLI option, but I'm not going to die on this hill :P

https://github.com/radicle-dev/radicle-client-services/pull/183/commits/8165c6c95a0ad2761d5fc3df0880167f5e749b95

PR #184, opened 2022-08-29
Improve project listing speed

adaszko:
From 30 seconds down to ~7 on my machine (`Aug 29 13:03:41.805  INFO request{method=GET uri=/v1/projects status=200 latency=6.635511333s}: radicle_http_api: Processed`).  99% of time is spent blocked on storage pool `get()`, so that's the bottleneck preventing further optimisation.


cloudhead (approved):
Nice!

PR #186, opened 2022-09-02
Log storage pool get() latency

adaszko:


http-api/src/lib.rs
```diff
@@ -103,10 +111,15 @@ impl Context {
...
-        self.pool
+        let start = Instant::now();
+        let result = self
+            .pool
             .get()
             .await
-            .map_err(|e| Error::Pool(e.to_string()))
+            .map_err(|e| Error::Pool(e.to_string()));
+        let milliseconds = saturating_u64_from_u128(start.elapsed().as_millis());
+        tracing::info!(milliseconds, "Got storage pool handle");
```
cloudhead, line 121: Isn't this going to be super noisy? Maybe we use `Trace` level?

http-api/src/lib.rs
```diff
@@ -103,10 +111,15 @@ impl Context {
...
     async fn storage(&self) -> Result<deadpool::managed::Object<Storage, InitError>, Error> {
-        self.pool
+        let start = Instant::now();
+        let result = self
+            .pool
             .get()
             .await
-            .map_err(|e| Error::Pool(e.to_string()))
+            .map_err(|e| Error::Pool(e.to_string()));
+        let milliseconds = saturating_u64_from_u128(start.elapsed().as_millis());
```
cloudhead, line 120: You can do this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=744c0d91f9ffb53157e6018bfbbbc7ea

cloudhead:
Sounds good!

PR #193, opened 2022-09-06
control: add track/untrack support

slack-coder:


node/src/test/tests.rs
```diff
@@ -167,6 +168,50 @@ fn test_inventory_sync() {
...
+    alice.command(Command::Untrack(proj_id.clone(), sender));
+    let policy_change = receiver
+        .recv()
+        .map_err(client::handle::Error::from)
+        .unwrap();
+    assert_eq!(true, policy_change);
+
+    // TODO: confirm inclusion in set of tracked peers.
+
+    // confirm internal state changes
```
cloudhead, line 205: No state is changing in this case though?

node/src/test/tests.rs
```diff
@@ -167,6 +168,50 @@ fn test_inventory_sync() {
     assert_eq!(a, b);
 }
 
+#[test]
+fn test_tracking() {
+    let rng = fastrand::Rng::new();
+    let mut alice = Peer::config(
+        "alice",
+        Config {
+            project_tracking: ProjectTracking::Allowed(HashSet::with_hasher(rng.clone().into())),
```
cloudhead, line 177: FYI you can just do `HashSet::default()` now, when you don't care about reproducibility

node/src/test/tests.rs
```diff
@@ -167,6 +168,50 @@ fn test_inventory_sync() {
...
+            ..Config::default()
+        },
+        [7, 7, 7, 7],
+        vec![],
+        MockStorage::empty(),
+        fastrand::Rng::new(),
+    );
+    let proj_id: identity::ProjId = test::arbitrary::gen(100);
```
cloudhead, line 185: The number passed to `gen` only really matters when generating arbitrary-sized data, but since project IDs are always the same size, I'd just pass `1` as a convention. Maybe `0` works too *shrug*

node/src/test/tests.rs
```diff
@@ -167,6 +168,50 @@ fn test_inventory_sync() {
...
+
+    let (sender, receiver) = chan::bounded(1);
+    alice.command(Command::Untrack(proj_id.clone(), sender));
+    let policy_change = receiver
+        .recv()
+        .map_err(client::handle::Error::from)
+        .unwrap();
+    assert_eq!(true, policy_change);
+
+    // TODO: confirm inclusion in set of tracked peers.
```
cloudhead, line 203: You could just do `assert!(alice.config.is_tracking(proj_id))`

cloudhead (commented):
Very nice! Commented on a few minor things that could be improved.

cloudhead:
(Merged)

PR #194, opened 2022-09-09
http-api: Set cache header for activity endpoint

rudolfs:
Allow caching activity responses for an hour.

Screenshot 2022-09-09 at 15 15 33

This might improve loading times for subsequent visits of a seed index page, e.g. https://app.radicle.xyz/seeds/maple.radicle.garden.


cloudhead (approved):
Nice!

PR #196, opened 2022-09-15
Use `radicle_surf` `Commit` struct instead of `radicle_source`

sebastinez:
This allows us to query parents of a commit and since radicle_source will eventually be deprecated, it prepares for it

Signed-off-by: Sebastian Martinez <me@sebastinez.dev>

### DIFF
--- a/http-api/src/commit.rs
+++ b/http-api/src/commit.rs
@@ -1,8 +1,8 @@
 use serde::{Deserialize, Serialize};
 
 use radicle_common::project::PeerInfo;
-use radicle_source::commit::Header;
-
+use radicle_surf::diff;
+use radicle_surf::vcs::git;
 #[derive(Serialize, Deserialize, Clone)]
 #[serde(rename_all = "kebab-case")]
 pub struct CommitsQueryString {
@@ -16,16 +16,16 @@ pub struct CommitsQueryString {
 
 #[derive(Serialize)]
 pub struct CommitTeaser {
-    pub header: Header,
+    pub header: git::Commit,
     pub context: CommitContext,
 }
 
 #[derive(Serialize)]
-pub struct Commit {
-    pub header: Header,
-    pub stats: radicle_source::commit::Stats,
-    pub diff: radicle_surf::diff::Diff,
-    pub branches: Vec<radicle_source::Branch>,
+pub struct Header {
+    pub header: git::Commit,
+    pub stats: diff::Stats,
+    pub diff: diff::Diff,
+    pub branches: Vec<git::BranchName>,
     pub context: CommitContext,
 }
 
--- a/http-api/src/commit.rs
+++ b/http-api/src/commit.rs
@@ -21,7 +21,7 @@ pub struct CommitTeaser {
 }
 
 #[derive(Serialize)]
-pub struct Header {
+pub struct Commit {
     pub header: git::Commit,
     pub stats: diff::Stats,
     pub diff: diff::Diff,
--- a/http-api/src/v1/projects.rs
+++ b/http-api/src/v1/projects.rs
@@ -18,18 +18,17 @@ use librad::git::identities::{self, SomeIdentity};
 use librad::git::types::{Namespace, One, Reference, Single};
 use librad::git::Storage;
 use librad::git::Urn;
+use librad::git_ext::Oid;
 use librad::paths::Paths;
 use librad::PeerId;
 
 use radicle_common::cobs::{self, issue, patch, Store};
 use radicle_common::person;
-use radicle_source as source;
-use radicle_source::commit::Stats;
-use radicle_source::surf::vcs::git;
 use radicle_surf::diff;
+use radicle_surf::vcs::git;
 
 use crate::axum_extra::{Path, Query};
-use crate::commit::{Commit, CommitContext, CommitTeaser, CommitsQueryString, Committer};
+use crate::commit::{CommitContext, CommitTeaser, CommitsQueryString, Committer, Header};
 use crate::project::{self, Info};
 use crate::{get_head_commit, Context, Error};
 
@@ -160,29 +159,47 @@ async fn project_root_handler(
 /// `GET /projects/:project/commits/:sha`
 async fn commit_handler(
     Extension(ctx): Extension<Context>,
-    Path((project, sha)): Path<(Urn, One)>,
+    Path((project, sha)): Path<(Urn, Oid)>,
 ) -> impl IntoResponse {
-    let reference = Reference::head(Namespace::from(project.clone()), None, sha.to_owned());
-    let commit = browse(reference, ctx.paths.clone(), |browser| {
-        let oid = browser.oid(&sha)?;
-        radicle_source::commit(browser, oid)
-    })
-    .await?;
-
     let projects = ctx.projects.read().await;
+
+    let repo = git::Repository::new(ctx.paths.git_dir()).map_err(Error::from)?;
+    let browser = git::Browser::new_with_namespace(
+        &repo,
+        &git::Namespace::try_from(project.encode_id().as_str())
+            .map_err(|_| Error::MissingNamespace)?,
+        *sha,
+    )
+    .map_err(Error::from)?;
+
+    let history = browser.get();
+    let commit = history.first();
+
+    let diff = if let Some(parent) = commit.parents.first() {
+        browser.diff(*parent, *sha)?
+    } else {
+        browser.initial_diff(*sha)?
+    };
+
     let fingerprints = projects.get(&project);
-    let fp = ctx.commit_ssh_fingerprint(&commit.header.sha1.to_string())?;
+    let fp = ctx.commit_ssh_fingerprint(&commit.id.to_string())?;
     let committer = if let (Some(fps), Some(fp)) = (fingerprints, fp) {
         fps.get(&fp).cloned().map(|peer| Committer { peer })
     } else {
         None
     };
 
-    let response = Commit {
-        header: commit.header,
-        diff: commit.diff,
-        stats: commit.stats,
-        branches: commit.branches,
+    let branches = browser
+        .revision_branches(*sha)?
+        .into_iter()
+        .map(|b| b.name)
+        .collect::<Vec<git::BranchName>>();
+
+    let response = Header {
+        header: commit.clone(),
+        stats: diff.stats(),
+        diff,
+        branches,
         context: CommitContext { committer },
     };
 
@@ -190,7 +207,7 @@ async fn commit_handler(
 }
 
 /// Get project commit range.
-/// `GET /projects/:project/commits?from=<sha>`
+/// `GET /projects/:project/commits?parent=<sha>`
 async fn history_handler(
     Extension(ctx): Extension<Context>,
     Path(project): Path<Urn>,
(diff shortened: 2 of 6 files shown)


sebastinez:
This PR relies on the implementation of https://github.com/radicle-dev/radicle-surf/pull/208
And since `radicle_surf` will eventually be migrated to `radicle-git` would have to be revisited

http-api/src/commit.rs
```diff
@@ -16,16 +15,25 @@ pub struct CommitsQueryString {
...
 #[derive(Serialize)]
 pub struct CommitTeaser {
-    pub header: Header,
+    pub header: git::Commit,
     pub context: CommitContext,
 }
 
+/// Commit statistics.
+#[derive(Clone, Serialize)]
+pub struct CommitStats {
```
FintanH, line 24: Wouldn't a commit only have stats compared to another commit? Or do you always calculate this based on the parent? :thinking:

http-api/src/v1/projects.rs
```diff
@@ -149,35 +149,93 @@ async fn commit_handler(
...
+    let diff = if let Some(parent) = commit.parents.first() {
+        browser.diff(*parent, sha)?
+    } else {
+        browser.initial_diff(sha)?
+    };
+
+    let mut deletions = 0;
+    let mut additions = 0;
+
+    for file in &diff.modified {
```
FintanH, line 177: Something you could do to make this function a little bit terser is to write an impl:
```rust
impl From<&Diff> for CommitStats {
  fn from(diff: &Diff) -> CommitStats {
    let mut deletions = 0;
    let mut additions = 0;
    /* these for loops in here */
    Self { additions, deletions }
  }
}
```

Then here, you would replace alllll this code with:
```rust
let stats = CommitStats::from(&diff);
```

sebastinez: Interesting that you say that, I think we can do even better.
We could add all this to `radicle_surf::diff::Diff` and then we only would need `Diff::stats()` for [`Changeset`](https://github.com/radicle-dev/radicle-client-services/blob/c3212beb1ce198ed31d03d7bd4b09c153900e118/http-api/src/v1/projects.rs#L685)  and for this handler. wdyt?

FintanH: Ya, that works! :) What is a `Changeset`? I'm curious as to why it has a `Diff` and a `Vec<Commit>`.

sebastinez: It is a wrapper around `Diff` `Stats` and `Vec<Commit>` have everything patch related in one object.
I'd have to think about if we can maybe, by extracting the `stats` function, into its elements.

FintanH: > It is a wrapper around `Diff` `Stats` and `Vec<Commit>` to have everything patch related in one struct.

Ah, I see :)

Also, I just came across this too https://docs.rs/git2/0.15.0/git2/struct.Diff.html#method.stats

sebastinez: Nice! Would it be possible, to remove our `diff::Diff` implementation and use the `git2::Diff` implementation?
Or maybe add some `TryFrom` trait between the two types?

sebastinez: Ok just refactored the `stats` retrieval out [7b2c38a](https://github.com/radicle-dev/radicle-client-services/pull/196/commits/7b2c38a40245b6c121068547b1d6a1b22b84ee1c)
and moved it to `radicle_surf` https://github.com/radicle-dev/radicle-git/pull/11

http-api/src/commit.rs
sebastinez: Yeah we only refer to these stats compared to another commit, with [7b2c38a](https://github.com/radicle-dev/radicle-client-services/pull/196/commits/7b2c38a40245b6c121068547b1d6a1b22b84ee1c) it's clearer since we only use it in the context of a `radicle_surf::diff::Diff`

http-api/src/v1/projects.rs
FintanH: > Nice! Would it be possible, to remove our `diff::Diff` implementation and use the `git2::Diff` implementation?
> Or maybe add some `TryFrom` trait between the two types?

There is a `TryFrom` :sweat_smile: https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/src/diff/git.rs#L86

sebastinez: Nice, but was more thinking along the lines of directly using `git2::Diff` but it seems like to have our own `diff::Diff` allows us some more space for implementation.

FintanH:
Is https://github.com/radicle-dev/radicle-git/pull/10 compatible with this change, btw? They're kind of fighting the same battles :)

http-api/src/v1/projects.rs
FintanH: Ya, this is true :) We'll have to see what happens when we look at the `radicle-surf` re-design.

sebastinez:
> Is [radicle-dev/radicle-git#10](https://github.com/radicle-dev/radicle-git/pull/10) compatible with this change, btw? They're kind of fighting the same battles :)

@FintanH I just converted this PR to draft state, since they are many moving parts currently and want the dust to settle a bit 😅

FintanH:
> @FintanH I just converted this PR to draft state, since they are many moving parts currently and want the dust to settle a bit sweat_smile

Would be good to get your eyes on those changes as we make them :)

http-api/src/v1/projects.rs
```diff
@@ -205,50 +222,53 @@ async fn history_handler(
...
+    let repo = git::Repository::new(ctx.paths.git_dir()).map_err(Error::from)?;
+    let browser = git::Browser::new_with_namespace(
+        &repo,
+        &git::Namespace::try_from(project.encode_id().as_str())
+            .map_err(|_| Error::MissingNamespace)?,
+        sha,
+    )
+    .map_err(Error::from)?;
+    let commits = browser.get();
+    let stats = browser.get_stats()?;
```
cloudhead, line 252: The problem with using surf here is that it always fetches the entire commit history, so this will be really slow for large repos.

http-api/src/v1/projects.rs
```diff
@@ -781,8 +752,8 @@ fn resolve_revisions(
 
             let mut changeset: Option<Changeset> = None;
 
-            if let (Ok(commits), Ok(diff), true) = (
-                radicle_source::commits::<PeerId>(browser, None),
+            if let (commits, Ok(diff), true) = (
+                browser.get().iter().cloned().collect::<Vec<git::Commit>>(),
```
cloudhead, line 756: Same here, this gets the full history (unless you changed it).

FintanH: Just an idea I wanna throw out there while the surf redesign happens and helps with those performance issues -- perhaps you could model some of this stuff with traits. That way you can define these functions generically and supply your own implementation. Then when surf looks better you just have to change what implements the trait :)

sebastinez: Hey @FintanH thanks for your input and help over Discord, I want this PR to be merged soon, so I'll propose to keep from adding this abstraction for another PR, and call `radicle_surf` directly.

http-api/src/v1/projects.rs
sebastinez: As talked over we keep fetching for now the entire commit history, but since we are paginating we at least do not return all the commits.

http-api/src/v1/projects.rs
sebastinez: As talked over we keep fetching for now the entire commit history, but since we are paginating we at least do not return all the commits.

http-api/src/v1/projects.rs
```diff
@@ -160,37 +159,56 @@ async fn project_root_handler(
...
     };
 
-    let response = Commit {
-        header: commit.header,
-        diff: commit.diff,
-        stats: commit.stats,
-        branches: commit.branches,
+    let branches = browser
+        .revision_branches(*sha)?
+        .iter()
```
FintanH, line 194: You can `into_iter` to take ownership of the data and you don't need the `cloned` then.

http-api/src/v1/projects.rs
```diff
@@ -791,11 +763,10 @@ fn resolve_revisions(
                 // Iterates over commits headers and retrieves each commit details until it gets to the head of the default branch
                 // If radicle_source::commit returns a None the commit won't be collected.
                 let commits = commits
-                    .headers
                     .iter()
```
FintanH, line 766: Same again, you might be able to `into_iter` here.

FintanH (commented):
Looks good to me

http-api/src/v1/projects.rs
sebastinez: Done

http-api/src/v1/projects.rs
sebastinez: This would be nice, but we would need to implement `IntoIter` for `History` to be able to do this.

sebastinez: Will do a comment somewhere on `radicle_git` to take that into account on the redesign

http-api/src/commit.rs
```diff
@@ -16,16 +16,16 @@ pub struct CommitsQueryString {
...
     pub context: CommitContext,
 }
 
 #[derive(Serialize)]
-pub struct Commit {
-    pub header: Header,
-    pub stats: radicle_source::commit::Stats,
-    pub diff: radicle_surf::diff::Diff,
-    pub branches: Vec<radicle_source::Branch>,
+pub struct Header {
```
cloudhead, line 24: Since `git::Commit` is qualified, we can call this simply `Commit`, as it includes more than just the header.

cloudhead:
Does this require any changes on the frontend?

sebastinez:
> Does this require any changes on the frontend?

Sorry for the delay was down in router land 😅.
Yeah we would need to merge https://github.com/radicle-dev/radicle-interface/pull/396

sebastinez:
For review a diff on how the commit header output would change

```diff
diff --git a/1.json b/2.json
index 3c56fcd..bee84f2 100644
--- a/1.json
+++ b/2.json
@@ -1,16 +1,21 @@
 {
   "header": {
-    "sha1": "47596419a5fcc9c360d5fee90e2fddbe733e1386",
+    "id": "47596419a5fcc9c360d5fee90e2fddbe733e1386",
     "author": {
       "name": "ray.wu",
-      "email": "wgq.quan@gmail.com"
+      "email": "wgq.quan@gmail.com",
+      "time": 1608263931
     },
     "summary": "optimize",
     "description": "",
+    "message": "optimize\n",
     "committer": {
       "name": "ray.wu",
-      "email": "wgq.quan@gmail.com"
+      "email": "wgq.quan@gmail.com",
+      "time": 1608263931
     },
-    "committerTime": 1608263931
+    "parents": [
+      "e10d0070001b3c40e38437bf58329473f0674221"
+    ]
   }
 }
```

cloudhead:
Interesting, so the message contains the summary in it. But then what is description? Seems like there's one field too many there.

sebastinez:
> Interesting, so the message contains the summary in it. But then what is description? Seems like there's one field too many there.

Yeah `message` is part of the `surf::Commit` while `description` is a getter that strips the summary.
Don't think the `message` part will come in much handy, would be a change to `radicle-git` to remove this.
Maybe at one point we want to show a raw format of a commit, and then `message` could come in handy 🤔

sebastinez:
Hmm ok I took this morning to read over all the design intentions of the `radicle-surf` redesign.
I'm thinking we should maybe close this PR and take another intent at this, once the redesign is finished, since it's not something urgent.
wdyt @cloudhead?

cloudhead:
Yeah sounds good.

PR #203, opened 2022-10-03
Return 404 on bad paths

adaszko:
Requests like `GET /favicon.ico/` currently result in 5xx error tripping alerts spuriously.  This PR makes it so that 404 is returned instead.

### DIFF
--- a/git-server/src/error.rs
+++ b/git-server/src/error.rs
@@ -150,6 +150,7 @@ impl Error {
             Error::Unauthorized(_) => http::StatusCode::UNAUTHORIZED,
             Error::KeyMismatch { .. } => http::StatusCode::UNAUTHORIZED,
             Error::AliasNotFound => http::StatusCode::NOT_FOUND,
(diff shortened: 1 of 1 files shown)


git-server/src/lib.rs
```diff
@@ -403,7 +394,42 @@ async fn git_handler(
...
+    .await
+    {
+        Ok((status, headers, body)) => (status, headers, body),
+        Err(e) => return e.into_response(),
+    };
+
+    match axum_resp_from_status_headers_body_triple(status, headers, body) {
+        Ok(resp) => resp,
+        Err(e) => e.into_response(),
+    }
```
cloudhead, line 432: What was the reason to move this out?

adaszko: Every occurrence of `?` would otherwise have to be replaced with `match […] { …, Err(e) => return e.into_response() }`.  If it’s factored out, it’s still possible to use `?` and handle all the errors in a single place.

PR #206, opened 2022-10-10
http-api: Change the iterator to not count failed projects

### LINKED ISSUE #205: http-api: Amount of projects fetched from /v1/projects wrong
It seems that by filtering out projects that return an `Err` we do not provide the amount of projects asked from the endpoint.

This probable happens when obtaining their `Info`, after doing the pagination with `skip` and `take`.

My idea is that we should get and validate the projects info during the iterator, so we always can return the asked amount.

To reproduce the issue, try for example https://maple.radicle.garden:8777/v1/projects?per-page=10&page=2 and I obtained 6 instead of 10 projects.

sebastinez:
To avoid returning a lesser amount of projects then requested, we get the project info during the iterator and paginate after filtering for invalid projects.

This removes the ability to fetch projects in parallel, but fixes the issue of returning less projects then requested for.

PS. If someone has an idea to fetch in parallel during the iterator, please feel free to commit to this PR or lmk how you would do it.

Closes #205

### DIFF
--- a/http-api/src/v1/projects.rs
+++ b/http-api/src/v1/projects.rs
@@ -106,43 +106,80 @@ async fn get_projects_info(
     let per_page = per_page.unwrap_or(10);
 
     let storage = ctx.storage().await?;
-    let projects: Vec<Project> = identities::any::list(storage.read_only())?
-        .map(|res| match res {
-            Ok(id) => Ok(id),
-            Err(err) => Err(Error::from(err)),
-        })
-        .filter_map(|id_result| match id_result {
-            Ok(SomeIdentity::Project(project)) => Some(Ok(project)),
-            Err(err) => Some(Err(err)),
-            _ => None,
-        })
-        .skip(page * per_page)
-        .take(per_page)
-        .collect::<Result<Vec<Project>, Error>>()?;
 
-    let pending_futures = {
-        let mut futures: Vec<tokio::task::JoinHandle<_>> = Vec::with_capacity(projects.len());
-        for id in &projects {
-            let id = id.clone();
-            let ctx = ctx.clone();
-            let future = tokio::task::spawn(get_project_info(ctx, id));
-            futures.push(future);
+    let mut infos: Vec<Info> = Vec::with_capacity(page);
+    let mut num_broken_projects = 0;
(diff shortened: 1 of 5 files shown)


http-api/src/v1/projects.rs
```diff
@@ -106,44 +100,23 @@ async fn get_projects_info(
...
-        .map(|res| match res {
-            Ok(id) => Ok(id),
-            Err(err) => Err(Error::from(err)),
-        })
+    let projects = identities::any::list(storage.read_only())?
         .filter_map(|id_result| match id_result {
-            Ok(SomeIdentity::Project(project)) => Some(Ok(project)),
-            Err(err) => Some(Err(err)),
+            Ok(SomeIdentity::Project(project)) => {
+                if let Ok(info) = get_project_info_sync(ctx.clone(), project, storage.as_ref()) {
```
adaszko, line 106: Hey, you have removed the `spawn_blocking()` stuff.  `get_project_info_sync()` is a pretty expensive operation (and a blocking one) and it tends to block processing of all HTTP requests if called in a blocking manner.  `spawn_blocking()` is there to keep on processing other requests while we wait for results of `get_project_info_sync()`

adaszko (changes requested):

http-api/src/v1/projects.rs
sebastinez: Ok, that makes sense.. Would it be possible to block the iterator somehow to wait for results of `get_project_info_sync()` so we can paginate over a collection of valid projects?

adaszko: I’m not sure what you mean by blocking the iterator exactly but one way fix the issue is to wrap the entire logic (with the `spawn_blocking()` and `join_all()`) in a `loop {}` that breaks when a sufficient number of project infos has been collected (or there isn’t any more projects left).  In case it’s not clear, any iterator can be advanced one step simply by calling its `next()` method:

```Rust
loop {
  let res = match identities::any::list(storage.read_only())? {
    Some(res) => res,
    None => break,
  };
  […]
}
```

adaszko (approved):

sebastinez:
Closing this in favour of a new PR on `radicle-dev/heartwood`

PR #209, opened 2022-10-10
http-api: Cast project aliases to lowercase

### LINKED ISSUE #201: http-api: Allow case-insensitive project alias queries
Ok have thought about this a bit, and while we already do case-insensitive searches against ENS profiles (with room for improvement), I'm not sure that this is practical for projects, since we don't enforce users to only use lower case project names.

sebastinez:
Closes #201


cloudhead:
Don't we also need to lowercase the requests?

sebastinez:
Closing this in favour of a new PR on `radicle-dev/heartwood`

